PR #71, opened 2016-03-16
Set sleep time between pipelines

georgiana-b:
Add sleep_time as part of batch_options.

### DIFF
--- a/goodtables/pipeline/batch.py
+++ b/goodtables/pipeline/batch.py
@@ -20,9 +20,7 @@ class Batch(object):
     * `source_type`: 'csv' or 'datapackage'
     * `data_key`: The key for the data (only used when source_type is 'csv')
     * `schema_key`: The key for the schema (only used when source_type is 'csv').
-    * `batch_options`: Options to pass to each batch instance.
-        * Set a sleep time (in seconds) between pipelines
-            * e.g., {'sleep_time': 0.5}
+    * `sleep`:  The time (in seconds) to wait beteween pipelines
     * `pipeline_options`: Options to pass to each pipeline instance.
     * `post_task`: Task handler to run after all pipelines have run.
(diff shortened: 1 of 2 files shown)


goodtables/pipeline/batch.py
```diff
@@ -19,6 +20,9 @@ class Batch(object):
     * `source_type`: 'csv' or 'datapackage'
     * `data_key`: The key for the data (only used when source_type is 'csv')
     * `schema_key`: The key for the schema (only used when source_type is 'csv').
+    * `batch_options`: Options to pass to each batch instance.
```
pwalsh: All options are "batch_options" here, except the `pipeline_options` which is an object to pass to each pipeline. So, adding a generic `batch_options` does not follow the API pattern here. It would be better to have a keyword argument like `rate` or `frequency` or `sleep`, with a default value of `1`.

So, the kwarg would be (for example) `rate=1`. 

Then, the try/except for the rate is also a bit non-idiomatic. Once you set a sensible default, like `rate=1`, then you should do the checking in the `__init__` and not let bad values exist, only to be discovered in the `run` method.

So, something like this in `__init__`, to catch the error immediately when passed:

```

if not isinstance(rate, (int, float)):
    raise ValueError('Received non int or float for the rate argument.')
self.rate = rate
```

pwalsh:
As well as the above comments, the test added actually fails:

https://travis-ci.org/frictionlessdata/goodtables/jobs/116277552

georgiana-b:
I am currently trying to figure out why the tests executed locally fail differently than these. Please ignore what I'm doing.

georgiana-b:
This has been included in the [`feature\include_datatable_errors_in_report`](https://github.com/frictionlessdata/goodtables/pull/78) and thus will be closed.

PR #74, opened 2016-03-24
Dialect Sniffing and Bytes support

akariv:


goodtables/datatable/datatable.py
```diff
@@ -118,6 +118,13 @@ def to_textstream(self, data_source):
 
                 return textstream
 
+        elif isinstance(data_source, compat.bytes):
+            self.encoding = self._detect_stream_encoding(data_source)
+            print(self.encoding)
```
pwalsh: can we remove the print function.

goodtables/compat.py
```diff
@@ -43,13 +43,21 @@ def iterenc_utf8(data):
     from urllib import parse
     from urllib.request import urlopen
     from urllib.error import HTTPError
-    csv_reader = csv.reader
```
pwalsh: As this change is only for Python 3 code, we are breaking Py2/3 compatibility. No?

akariv: Yes, I'll add that same functionality for Py2

akariv:
Okay @pwalsh , fixed the tests and updated the PR.
Note that some changes were made to adapt to the fact that the `required` constraint was made to be false by default (https://github.com/frictionlessdata/jsontableschema-py/commit/45ebff840eea74a5ca20f80a02a159bacad5330d)

pwalsh:
@akariv still a bunch of failing tests.

akariv:
Probably because it's using a cached version of `jsontableschema-py` in its
dependencies.
Can you try and invalidate the cache?

pwalsh:
ok, ready to merge

akariv:
@pwalsh I don't have the creds, can you merge?
Also we need to publish to pypi for goodtables.io to build correctly

PR #76, opened 2016-03-27
Remove indexes in reports

trickvi:
Remove index mentions from error messages in reports by simplifying messages.

Also added a fix to a bug I noticed while working on this.


pwalsh:
Thanks @trickvi

PR #77, opened 2016-03-29
Hotfix/check contraints before unique

georgiana-b:
Added forgotten check in `schema processor` and updated the test.


pwalsh:
thanks.

PR #83, opened 2016-04-15
Feature/include xls and csv errors in report

georgiana-b:
Solves #84.


goodtables/compat.py
```diff
@@ -45,22 +46,28 @@ def iterenc_utf8(data):
 
         iter = chain(first_lines, data)
         iter = iterenc_utf8(iter)
-        reader = csv.reader(iter, dialect=dialect, **kwargs)
-        for row in reader:
-            yield [str(cell, 'utf-8') for cell in row]
+        csv.field_size_limit(20000000)
```
pwalsh: @georgiana-b you cant add a file size limit here. We need to be able to accept huge files, so this is not a solution.

pwalsh:
@akariv can you please review this PR when you can.

goodtables/compat.py
pwalsh: right, my mistake, it is a field size limit.

georgiana-b: It's a _field_ size limit that used to have the default value 131072 which wasn't enough for some of the files so I changed it to a number unlikely to be reached.

akariv:
@pwalsh this looks fine to me.

pwalsh:
@akariv thanks

PR #89, opened 2016-04-30
Improve excel error handling

georgiana-b:
Stop excel error handling from interfering with HTTP errors and make it handle strings that are not paths gracefully.


pwalsh:
published in gt 0.7.1

PR #90, opened 2016-04-30
Quote composed urls separately

georgiana-b:
[urllib](https://docs.python.org/2/library/urllib.html) doesn't support composed URLs like `http://webarchive.nationalarchives.gov.uk/+/http://www.nio.gov.uk/transaction_spend_data_august_10_northern_ireland_office.xls`. Because it doesn't recognize there are 2 URLs in that string, it's quoting the `/+/http://` part, which results in those URLs returning 500.  This is a workaround to that issue.

### DIFF
--- a/goodtables/utilities/helpers.py
+++ b/goodtables/utilities/helpers.py
@@ -208,7 +208,11 @@ def make_valid_url(url):
     Args:
         * `url`: a url string
     """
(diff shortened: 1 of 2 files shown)


pwalsh:
published in gt 0.7.1

PR #96, opened 2016-05-07
Feature/validate pipeline options

georgiana-b:
Validate pipeline options and processors parameters.

### DIFF
--- a/goodtables/pipeline/pipeline.py
+++ b/goodtables/pipeline/pipeline.py
@@ -7,9 +7,11 @@ from __future__ import unicode_literals
 import io
 import importlib
 import tellme
+import inspect
 from ..utilities import csv_dialect, helpers
 from .. import datatable
 from .. import exceptions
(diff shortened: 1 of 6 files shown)


goodtables/pipeline/pipeline.py
```diff
@@ -144,6 +145,16 @@ def get_pipeline(self):
...
         return pipeline
 
+    def validate_processors(self):
+        """Validates the processors parameter"""
+
+        if not isinstance(self.processors, (list, tuple, set)):
+            msg = 'The \'processors\' argument must be a list, tuple or set.'
+            raise exceptions.InvalidPipelineOptions(msg)
+
+        for processor_name in self.processors:
```
georgiana-b: I chose to iterate though all the processors and let the `resolve_processor` raise the error if a processor in invalid. If you have a better idea of how to handle this, please share.

Issue #98, opened 2016-05-10
CSV incorectly split into columns


georgiana-b:
[This CSV](http://goodtables.okfnlabs.org/reports?data_url=https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/252886/July_2013_GPC_Transparency_Data.csv&format=CSV&encoding=&schema_url=https://raw.githubusercontent.com/okfn/tabular-validator/master/examples/hmt/spend-publishing-schema.json) seems to be incorrectly parsed. Investigate the issue.

roll:
@georgiana-b 
Do we still have an incorrect behaviour here?

http://goodtables.okfnlabs.org/reports?data_url=https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/252886/July_2013_GPC_Transparency_Data.csv&format=CSV&encoding=&schema_url=https://raw.githubusercontent.com/okfn/tabular-validator/master/examples/hmt/spend-publishing-schema.json

pwalsh:
@roll if this acts correctly in v1 I am happy to close this.

roll:
```bash
$ goodtables table https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/252886/July_2013_GPC_Transparency_Data.csv
DATASET
=======
{'error-count': 9, 'table-count': 1, 'time': 0.988, 'valid': False}

TABLE [1]
=========
{'error-count': 9,
 'headers': ['Transaction reference number',
             'Posting Date',
             'Merchant Name',
             'Final Transaction Amount',
             'Description',
             '',
             '',
             '',
             ''],
 'row-count': 26,
 'time': 0.982,
 'valid': False}
---------
[-,6] [blank-header] Header in column 6 is blank
[-,7] [blank-header] Header in column 7 is blank
[-,8] [blank-header] Header in column 8 is blank
[-,9] [blank-header] Header in column 9 is blank
[-,7] [duplicate-header] Header in column 7 is duplicated to header in column(s) 6
[-,8] [duplicate-header] Header in column 8 is duplicated to header in column(s) 6, 7
[-,9] [duplicate-header] Header in column 9 is duplicated to header in column(s) 6, 7, 8
[24,-] [blank-row] Row 24 is completely blank
[25,-] [blank-row] Row 25 is completely blank
```

roll:
I don't see any problems with this file.

abhiravbansal:
Hi,

[...]

PR #104, opened 2016-05-19
Improve CSV dialect detection

### LINKED ISSUE #100: CSV is not being parsed correctly and GT crashes analysing it
Trying to validate this file (for further import on OS) but don't see what's wrong in it and GT doesn't help.

[PEF2016_AC01.csv.zip](https://github.com/frictionlessdata/goodtables/files/268161/PEF2016_AC01.csv.zip)

georgiana-b:
It fixes #100, #98.
Since the CSV dialect detection is statistical, having a fixed value isn't suitable for all cases and the detection fails for very short or very long files.
This implementation starts sampling from 2 rows and continues to add until it finds a delimiter that is not (entirely) alaphanumeric or space. It is also the lightest I could come up with. If you have a better idea, please share.

### DIFF
--- a/goodtables/compat.py
+++ b/goodtables/compat.py
@@ -8,7 +8,8 @@ import sys
 import tempfile
 import io
 import csv
-from itertools import islice, chain
+import re
+from itertools import islice
 from goodtables import exceptions
 
 _ver = sys.version_info
@@ -34,18 +35,18 @@ if is_py2:
         """Read text stream (unicode on Py2.7) as CSV."""
 
         first_lines = list(islice(data, header_index, header_index + 2))
-        try:
-            dialect = csv.Sniffer().sniff(''.join(first_lines))
-            dialect.delimiter = dialect.delimiter.encode('utf-8')
-            dialect.quotechar = dialect.quotechar.encode('utf-8')
-        except csv.Error:
-            dialect = csv.excel
-
+        dialect = detect_dialect(first_lines)
+        
+        while not re.findall('[^\w ]', dialect.delimiter): 
+            first_lines.extend(list(data.readline()))
+            dialect = detect_dialect(first_lines)
+            
         def iterenc_utf8(data):
             for line in data:
                 yield line.encode('utf-8')
 
-        iter = chain(first_lines, data)
+        data.seek(0)
+        iter = data
         iter = iterenc_utf8(iter)
         csv.field_size_limit(20000000)
         try:
@@ -54,6 +55,19 @@ if is_py2:
                 yield [str(cell, 'utf-8') for cell in row]
         except TypeError as e:
             raise exceptions.DataSourceMalformatedError(msg=e.args[0], file_format='csv')
+    
+                
+        
+    def detect_dialect(sample):
+        """Detect delimiter and quote chars"""
+
+        try:
+            dialect = csv.Sniffer().sniff(''.join(sample))
+            dialect.delimiter = dialect.delimiter.encode('utf-8')
+            dialect.quotechar = dialect.quotechar.encode('utf-8')
+            return dialect
+        except csv.Error:
+            return csv.excel
 
 
 elif is_py3:
@@ -74,18 +88,28 @@ elif is_py3:
                 yield line
 
         first_lines = list(islice(data, header_index, header_index + 2))
-        data.seek(0)
-        try:
-            dialect = csv.Sniffer().sniff(''.join(first_lines))
-        except csv.Error:
-            dialect = csv.excel
+        dialect = detect_dialect(first_lines)
 
+        while not re.findall('[^\w ]', dialect.delimiter): 
+            first_lines.extend(list(data.readline()))
+            dialect = detect_dialect(first_lines)
+
+        data.seek(0)
         iter = line_iterator(data)
         csv.field_size_limit(20000000)
         try:
             return csv.reader(iter, dialect, **kwargs)
         except TypeError as e:
             raise exceptions.DataSourceMalformatedError(msg=e.args[0], file_format='csv')
+        
+    def detect_dialect(sample):
+        """Detect delimiter and quote chars"""
+
+        try:
+            dialect = csv.Sniffer().sniff(''.join(sample))
+            return dialect
+        except csv.Error:
+            return csv.excel
 
 
 def to_bytes(textstring):
--- a/goodtables/compat.py
+++ b/goodtables/compat.py
@@ -37,7 +37,7 @@ if is_py2:
         first_lines = list(islice(data, header_index, header_index + 2))
         dialect = detect_dialect(first_lines)
         
-        while not re.findall('[^\w ]', dialect.delimiter): 
+        while re.findall('[\w ]', dialect.delimiter):
             first_lines.append(data.readline())
             dialect = detect_dialect(first_lines)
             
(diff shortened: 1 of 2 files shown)


pwalsh:
@akariv WDYT? I say we merge and give it a try.

goodtables/compat.py
```diff
@@ -30,22 +31,22 @@
...
-            dialect.delimiter = dialect.delimiter.encode('utf-8')
-            dialect.quotechar = dialect.quotechar.encode('utf-8')
-        except csv.Error:
-            dialect = csv.excel
-
+        first_lines = list(islice(data, header_index, header_index + 2))
+        dialect = detect_dialect(first_lines)
+        
+        while not re.findall('[^\w ]', dialect.delimiter): 
+            first_lines.extend(list(data.readline()))
```
akariv: why not use `append(` instead of `extend(list(`?

goodtables/compat.py
```diff
@@ -30,22 +31,22 @@
...
+            dialect = detect_dialect(first_lines)
+            
         def iterenc_utf8(data):
             for line in data:
                 yield line.encode('utf-8')
 
-        iter = chain(first_lines, data)
+        data.seek(0)
```
akariv: why do you assume that data is seekable?
why was this change needed?
if it was, why keep first_lines?

pwalsh: the DataTable class ensures that what is passed here is a seekable stream. a bit unclear, I know.

goodtables/compat.py
```diff
@@ -30,22 +31,22 @@
...
-            dialect = csv.Sniffer().sniff(''.join(first_lines))
-            dialect.delimiter = dialect.delimiter.encode('utf-8')
-            dialect.quotechar = dialect.quotechar.encode('utf-8')
-        except csv.Error:
-            dialect = csv.excel
-
+        first_lines = list(islice(data, header_index, header_index + 2))
+        dialect = detect_dialect(first_lines)
+        
+        while not re.findall('[^\w ]', dialect.delimiter):
```
akariv: better to use [`isspace`](https://docs.python.org/2/library/stdtypes.html#str.isspace) for this, no?

goodtables/compat.py
```diff
@@ -67,23 +81,35 @@ def iterenc_utf8(data):
...
+    def csv_reader(data, header_index, **kwargs):
 
         def line_iterator(data):
             for line in data:
                 yield line
+
+        first_lines = list(islice(data, header_index, header_index + 2))
+        dialect = detect_dialect(first_lines)
+
+        while not re.findall('[^\w ]', dialect.delimiter):
```
akariv: see above

goodtables/compat.py
```diff
@@ -67,23 +81,35 @@ def iterenc_utf8(data):
...
 
         def line_iterator(data):
             for line in data:
                 yield line
+
+        first_lines = list(islice(data, header_index, header_index + 2))
+        dialect = detect_dialect(first_lines)
+
+        while not re.findall('[^\w ]', dialect.delimiter): 
+            first_lines.extend(list(data.readline()))
```
akariv: see above

goodtables/compat.py
```diff
@@ -67,23 +81,35 @@ def iterenc_utf8(data):
...
+            dialect = detect_dialect(first_lines)
+
+        data.seek(0)
```
akariv: see above

goodtables/compat.py
akariv: Yes, but who guarantees that we should be processing the CSV from position 0?
When I made this change (seek -> chain) it wasn't the case. It was because we were using two csv readers - one to read the headers and then passing the stream to another csv reader to read the data. In that case, seeking to 0 would re-read the headers.

Therefore, I wouldn't seek to 0 - we can imagine a 'comment reader' that reads in the CSV until there are no more 'comment lines' (I remember seeing an issue about that once), or another who reads the 'BOM' character in the beginning of the stream. Then these processors pass the CSV reader the stream in the correct position to start, which is not 0.

pwalsh: yes, correct. that is what `header_index` is for. there are other places in code that seek using it. @georgiana-b

akariv: Is there any reason not to keep `iter = chain(first_lines, data)` ?

goodtables/compat.py
georgiana-b: because appending a list to a list creates a 2 dimensional list: 
```
In [7]: test_list = [2,3,4]
In [8]: test_list.append([6])
In [9]: test_list
Out[9]: [2, 3, 4, [6]]

In [10]: test_list.extend([8])
In [11]: test_list
Out[11]: [2, 3, 4, [6], 8]
```

akariv: No, only if you append a list. But `data.readline()` is not a list.
```
In [5]: test_list = [2,3,4]
In [6]: test_list.append(6)
In [7]: test_list
Out[7]: [2, 3, 4, 6]
```

goodtables/compat.py
georgiana-b: Adam, functions in DataTable (and maybe other places) expect to start from 0: https://github.com/frictionlessdata/goodtables/blob/master/goodtables/datatable/datatable.py#L209. I started sampling from the `header_index` because it influences the accuracy of the detection but if I then use `chain`, it will only include the lines from `header_index` downwards which conflicts with what other functions expect and leads to incorrect results.

goodtables/compat.py
georgiana-b: `[^\w ]` matches anything that is not alphanumeric or whitespace.  `not str.isspace()` returns true for alphanumeric which is not ok (ex: `m` is not a valid delimiter).

goodtables/compat.py
georgiana-b: You're right, there's no reason to convert that to a list. I'll change it.

goodtables/compat.py
akariv: You're right. 
Still, running a regexp match is quite slow.
Matching like `delimiter.isalnum() || delimiter in [' ', '_']` is quicker and IMO more readable.

goodtables/compat.py
akariv: If this is an assumption then it's a very recent assumption.

See this code from less than 2 months ago, https://github.com/frictionlessdata/goodtables/blob/519f49cc0373eaf30dad25e744a7a3ad29a4a637/goodtables/datatable/datatable.py#L63
`csv_reader` is assumed here to *not* change the stream position.

goodtables/compat.py
georgiana-b: That leaves uncovered the case for multiple spaces (ex: a tab that is actually 4 spaces). I think it's better to have all cases covered than to be very fast in this case because that check won't execute many times.

akariv: I don't think the delimiter can be more than 1 character, but alright.
We can still remove the double negation, right?

georgiana-b: Right, I removed it.

goodtables/compat.py
georgiana-b: It's not an assumption. This is how that same function looks these days: https://github.com/frictionlessdata/goodtables/blob/master/goodtables/datatable/datatable.py#L63. I don't understand what's bothering about this part.

akariv: You're changing the behaviour of this function - a behaviour that was once _assumed_ to be the exact opposite.
I want to make sure you checked that no one else in the code base is also assuming that the reader will start from the current position of the stream. 

Coming to think of it, who else in the code base needs the lines before the headers? Why not return an iterator that begins with the header row?

pwalsh:
closing because GoodTables datatable is about to be replaced by [Tabulator](https://github.com/frictionlessdata/tabulator-py), rendering this out of date.

PR #137, opened 2016-10-27
Renamed profile to preset with simplified API

### LINKED ISSUE #124: profile naming and interface
# Overview

There are two questions:
- `profile` is also in use for datapackage profiles 
- `profile` interface could be updated

## Naming

TBD

## Interface

@pwalsh has wrote:

---

I'm not a fan of passing in data (errors and table in this case), modifying it, and then not returning that modified data. https://github.com/frictionlessdata/goodtables-py/blob/next-initial/goodtables/profiles/datapackage.py#L15

Based on the way profiles are used here, I don't see why the profile function needs errors and table passed in as arguments - they are always empty. Why not get rid of the empty list assignments and have

```python
errors, tables = profile_handler(source, **options)  # sidenote: handler because more meaningful than func?
```

---

The reason why now `profiles` gets predefined `errors` and `tables` - API consistency with checks (but that's true we could removed errors and tables from `profile` signature):

```python
profile(errors, tables, ...)
check(errors, columns, ...)
```

Also cc @amercader

roll:
- fixes #124


roll:
thanks

PR #145, opened 2016-11-01
Improved tests

### LINKED ISSUE #127: Improve test coverage
# Description

We need to test all edge-cases etc. Many prev version data examples are in `data` folder.

# Tasks

- [ ] improve test coverage
- [x] set coverage threshold in coverals to 90%

roll:
- fixes [partially] #127

---

This is not covering all current goodtables data examples and test cases. We need to address full coverage later in #127.


roll:
@pwalsh 
could you please review?

pwalsh:
@roll all good. Are we sure that it is not better to just put this stuff in `testsuite` and not keep all these files in here anymore?

roll:
@pwalsh 
I think there are a few things:
- we definitely should reduce amount of data examples here because many of them just a duplication and I suppose we lack some edge-case about csv dialects, broken encoding etc
- but still if we could provide complete test coverage (including high-level tests) on the library level - let's do it! We need great coverage on lib level to work further - accepts PRs, refactor etc. With instant feedback in dev and CI.
- so testsuite role here - to have tests which guarantee that our stack works good together (e.g. tabulator update -> tabulator tests good -> testsuite goodtables tests fail -> we've missed breaking change on tabulator level) and works after every master/PyPi change. In my opinion it doesn't mean we need just automatically move all high-level tests to testsuite - only if we can't have it on the lib level (like datapackage + jsontableschema-sql integration)

PR #151, opened 2016-11-03
Fixed order fields work

roll:
# Description

It fixes problem with field ordering if there are less fields then headers and vica versa.


roll:
@pwalsh 
could you please review

roll:
Thanks

Issue #105, opened 2016-08-12
Release version-1


roll:
# Overview

Tracking issue for https://github.com/frictionlessdata/goodtables-py/milestone/9

# Tasks

- [x] update implementation
- [x] release ALPHA
---
- [x] provide visual separation for values inside error messages (add quotes doing spec template formatting) - MOVED #163
- [x] expose package version
- [x] remove `alpha` from dependencies
- [x] remove deprecated code
- [x] improve test coverage - MOVED #127 
- [x] set coverage threshold in coverals to 90% - MOVED #127
- [x] DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() instead
- [x] Readme improvements:
  - [x] highlight in docs that schema checks is not enabled by default
  - [x] Couple of simple library examples at the top - the obvious stuff
    - [x] Validate data in a datapackage (all or just one resource)
    - [x] Validate data in a CSV with and without a JSON Table Schema
  - [x] Explain conceptually the two types of checking that the library does: Structure and Schema
    - [x] Give a CLI datapackage example
    - [x] Explain what `infer-fields` does ...
- [ ] release

pwalsh:
@roll looks like this is now done!

roll:
@pwalsh 
But we still on alpha. Let's track here final release (as for the other libs).

PR #200, opened 2017-07-12
Improve API and readme

### LINKED ISSUE #186: Improve user-facing API
From @pwalsh:

1. MOVED to #198
2. MOVED to #198
3. CLI and Python interfaces: The new version of goodtables breaks consistency between the CLI and Python interfaces. `Inspector` is a new concept compared to old versions, and is the primary interface in Python. In CLI, inspector does not feature as a concept, and instead we expose the Inspector as `goodtables`. This is more in line with the old API semantics. However, on CLI, the order of arguments is reversed, and the user needs to provide the preset before the data. This is pretty confusing for no obvious reason.

---

From @rufuspollock:

Small suggestion for making this awesome library even better 😄 

Couple of convenience functions we could add:

```python=
# validate a data package
goodtables.validate_datapackage(url_or_path)

# validate an individual tabular data file 
goodtables.validate(datafile, schema=None)

datafile = url_or_path_to_table_data_or_fileobj_or_row_iterator # would be happy to just support fileobj and row iterator
schema = schema_url_path_or_file_obj_or_dict=None)
```

### LINKED ISSUE #199: Detect datapackage preset
# Overview

We could automatically set preset to `datapackage` if we see that's a data package. So e.g. for CLI it will be:

```bash
$ goodtables datapackage.json
```

instead of

```bash
$ goodtables datapackage.json --preset datapackage
```

The same could be for the `validate` function. We use the same practice (detection) on `tabulator` level if there is no scheme/format/encoding/etc provided.

roll:
- fixes #186
  - added `validate` top-level function
  - made CLI nicer (no more `table`, `datapackage` prefixes in command line) 
- fixes #199 
  - now preset is detected automatically for the most cases e.g. see [CLI example](https://github.com/frictionlessdata/goodtables-py/tree/api-readme-improvements#cli)
- connects #105  
  - improved readme - https://github.com/frictionlessdata/goodtables-py/tree/api-readme-improvements#goodtables-py
  - many small improvements


roll:
@amercader 
Could you please review. I'm targeting this PR as the final before `v1` release. (cc @pwalsh @danfowler)

README - https://github.com/frictionlessdata/goodtables-py/tree/api-readme-improvements#goodtables-py

README.md
```diff
@@ -1,138 +1,192 @@
...
+report = validate('data.csv', checks='schema') # only schema checks
+report = validate('data.csv', checks={'bad-headers': False}) # exclude 'bad-headers'
+report = validate('data.csv', checks={'bad-headers': True}) # check only 'bad-headers'
+```
 
-inspector = Inspector()
-inspector.inspect('datapackage.json', preset='datapackage')
+By default a dataset will be validated against all available Data Quality Spec errors. Some checks could be not available for validation e.g. if schema is not provided only `structure` checks will be done.
+
+#### `validate(source, **options)`
```
amercader: While reading the README I found this API documentation really confusing. It took me a while to understand that these are all params for the `validate` function but that depending on the preset different params are accepted. Perhaps instead of `[Dataset:table]` it would make more sense to use `[preset:table]`?

roll: @amercader 
great feedback - I think may be just describe it in plain words:
`[Arguments - for `table` preset]` etc

Issue #198, opened 2017-07-10
Finalize custom presets/checks API and checks parameter for validate function


roll:
# Overview

For `v1` custom presets and checks are provisional API. As a part of work on `contrib.checks` (#191) we tests all edge cases and finalize those APIs. Also we have make a final decision on how `validate(checks=...)` parameter works. We should take into account that it also should be easy-to-use in CLI and `goodtables.yml`.

# Tasks

- [ ] learn from all edge cases on `contrib.checks`
- [ ] handle feedback on interface provided below
- [ ] discuss `validate(checks)`
- [ ] make changes if needed
- [ ] improve documentation/tutorials

---

From @pwalsh:

Some issues I encountered using goodtables v1 pre.

1. Checks: There is no documentation, and the semantics are not clear. Example: Does `column` refer to a cell in a row, or a column of cells? It must be the former as we are streaming rows, but it is confusing because a column is not a cell.
2. Register: The `check` (and `preset`) decorators are implemented from the `register` module. This suggests that functions decorated as such may get added to a registry of checks and presets. However, we "register" custom checks and presets on the `Inspector` constructor explicitly, So it is not clear to me why we need decorators, when we can just ensure that custom presets and checks implement the signature required to be a `preset` or a `check`. As a user, it feels like extra configuration that I don't need.

roll:
@roll:

- improved `validate(checks)` argument. Examples (will be documented as a part of #198):
  - `checks='spec'` - default, all spec checks
  - `checks='schema'` - only schema checks
  - `checks=['structure', 'extra-header', 'missing-header']` - list expanding `structure` group
  - `checks={'blank-row': False}` - cherry-pick check
  - `checks={'spec': True, 'blank-row': False}` - all spec checks but not `blank-row`
  - `checks={'schema': True, 'required-constraint': False}` - all schema checks but not `required-constraint`
  - `checks={'sequential-column': {'column': 'id'}}` - one contrib check with parameters
  - etc

---

@amercader:

I'm not totally sold on how the checks parameter is working right now TBH, there's just too many variants. For instance is this possible? checks={'sequential-column': True} If so what about parameters?

Perhaps is best to have explicit params for validate: include_checks, exclude_checks? Not sure, just wondering.

You've given more thought to this so definitely not blocking on my side.

roll:
> checks={'sequential-column': True}
[...]

Issue #191, opened 2017-05-23
Implement contrib.checks


pwalsh:
# Description

Users and potential users are consistently asking for more complex checks. 

We've always supported extensibility in checks, and imagined the type of things that could be implemented for more involved data integrity validation. However, it is clear that the value of goodtables would greatly increase if some such checks were built in to the library, not in the least so that they would eventually be accessible via [goodtables.io](https://goodtables.io).

I recommend that such checks should be shipped as a `contrib` module, in the spirit of [Django Contrib](https://github.com/django/django/tree/master/django/contrib). We could further namespace into groups, for discoverability reasons, but probably not necessary now.

Common requests:

- Check that a column of integers increments
- Declare constraint logic across multiple columns
- Check variation in a column from a mean

# Tasks

- [x] Agree on a design for a contrib module for checks
- [x] Decide on a set of initial checks we can ship based on current needs and feedback
- [x] Implement decided checks

pwalsh:
also see link to probes in https://github.com/frictionlessdata/goodtables-py/issues/10

pwalsh:
I think we will move on a `contrib.checks` module sooner, rather than later. We've had many signals in recent months on the utility of this, including from pilot partners.

Here is another of checks that a pilot partner has identified, with some comments from me in `[]`:

- Check for administrative variables that should not be in the data [means scanning cells against a black list of terms]
- Check for duplicate variables [we do duplicate rows, this is the same, but duplicates in columns, and therefore supported directly with `constraints.unique: true`]
- Check out of range values according to the code book [classic case. can be done based on hard range limits, or % of deviation, or some other ways I suppose]
- Look for any missing values not defined in code book [check values against a passed lookup table - similar implementation to the blacklist]
- Look for truncated numeric values or string variables [probably supported with core checks]
- Check variable and value labels are complete and make sense/are not truncated [need clarity]
- Sometimes logical checks: - make sure men or 90 years old women have not had babies! (usually done thoroughly by the survey owner) [checks across fields is coming up a lot. I guess what we need is an abstract to apply other checks across multiple fields, combined with IF OR and AND statements]
- Check derived variables make sense according to documentation [not clear]
- Disclosure risk (small frequency counts, outliers) [can be down with very basic statistical analysis over all columns]

roll:
@pwalsh 
@amercader 
@danfowler 
So which checks we're going to implement on the first iteration?
- easy choice to be the first check is `sequential-column` (help with name)
- ???

[...]

PR #202, opened 2017-07-19
Bootstraped contrib.checks and improved custom presets/checks

roll:
- connects #191
- connects #198

---

Changes:
- rebased on registry to simplify custom `presets/checks` API. No more `validate(custom_presets, custom_checks)` arguments. All `presets/checks` are available after registration with decorator.
- improved `validate(checks)` argument. Examples (will be documented as a part of #198):
  - `checks='spec'` - default, all spec checks
  - `checks='schema'` - only schema checks
  - `checks=['structure', 'extra-header', 'missing-header']` - list expanding `structure` group
  - `checks={'blank-row': False}` - cherry-pick check
  - `checks={'spec': True, 'blank-row': False}` - all spec checks but not `blank-row`
  - `checks={'schema': True, 'required-constraint': False}` - all schema checks but not `required-constraint`
  - `checks={'sequential-column': {'column': 'id'}}` - one contrib check with parameters
  - etc
- support class-based checks - https://github.com/frictionlessdata/goodtables-py/blob/custom-presets-checks-update/goodtables/checks/contrib/sequential_column.py
- implemented `sequential-column` check (as a `contrib.checks` POC):

Python:
```python
# User could just use contrib.check by its name (we will list them in readme)
report = validate('data.csv', checks={
    'sequential-column': {'column': 'id'},
})
```

goodtables.yml:
```yml
files: data.csv
settings:
  checks:
    sequential-column: 
      column: 'id'
```


goodtables/checks/contrib/sequential_column.py
```diff
@@ -0,0 +1,65 @@
...
+        for item in columns:
+            if self.__column in [item['number'], item['header']]:
+                column = item
+                break
+
+        # Check column
+        if not column:
+            return errors.append({
+                'code': 'sequential-column',
+                'message': 'Sequential column violation',
```
amercader: Perhaps `'message': 'Sequential column violation (column not found)'`?

amercader:
@roll great stuff

I'm not totally sold on how the `checks` parameter is working right now TBH, there's just too many variants. For instance is this possible? `checks={'sequential-column': True}` If so what about parameters?

Perhaps is best to have explicit params for `validate`: `include_checks`, `exclude_checks`? Not sure, just wondering.

You've given more thought to this so definitely not blocking on my side.

goodtables/inspector.py
```diff
@@ -29,26 +23,24 @@ class Inspector(object):
     # Public
 
     def __init__(self,
-                 checks='all',
-                 error_limit=1000,
-                 table_limit=10,
-                 row_limit=1000,
+                 checks='spec',
```
roll: Yes it's internal default. Also in readme Data Quality Spec is having an important place so I suppose user should understand what's going on calling `validate('data.csv')` (do all available **spec** checks).

goodtables/checks/duplicate_row.py
```diff
@@ -6,32 +6,38 @@
...
-                'code': 'duplicate-row',
-                'message': message,
-                'row-number': row_number,
-                'column-number': None,
-            })
-            # Clear columns
-            del columns[:]
-        references.append(row_number)
+@check('duplicate-row', type='structure', context='body')
+class DuplicateRow(object):
```
roll: Yes. I'll do it as a part of https://github.com/frictionlessdata/goodtables-py/issues/198. For now plan to finalize it based on `contrib.checks` implementation experience.

goodtables/checks/unique_constraint.py
```diff
@@ -11,23 +11,29 @@
...
+class UniqueConstraint(object):
+
+    # Public
+
+    def __init__(self, **options):
+        self.__row_indexes = {}
+
+    def check_row(self, errors, columns, row_number):
+        for column in columns:
+            if len(column) == 4:
```
roll: The same as above (#198). It will be a good amount of writing stuff))) Also going to add comments and split on blocks current checks to make it easy to learn from it. So this line should be absolutely unclear for anyone except me for now. I'll document it.

In short `column` could not have `column.value` or `column.header` if it's not healthy column. Check for len is kinda lazy so I'll updated this condition to make it self-explanatory.

roll:
@amercader 
Thanks. Updated readme and commented. Also I've moved `validate(checks)` discussion here - #198

goodtables/checks/contrib/sequential_column.py
roll: I've used it as a `contrib.checks` POC so I think it should be renamed to `sequential-integers` and all messages here will be rewritten to be more helpful.

roll: I mean `integers-increment` or something from #191

PR #203, opened 2017-08-02
Implemented contrib.checks and various improvements

### LINKED ISSUE #191: Implement contrib.checks

### LINKED ISSUE #198: Finalize custom presets/checks API and checks parameter for validate function

roll:
- fixes #191
- fixes #198

---

- implemented following`contrib.checks` (per #191):
  - `blacklisted-value`
  - `deviated-value`
  - `sequential-value`
  - `truncated-value`
  - `custom-constraint`
- renamed `column` to `cell` in checks (per #198)
- improved readability of all builtin checks (more comments etc)
- simplified `checks` argument -> `checks` and `skip_checks` (both lists) (per #198) 
- removed `provisional API` mark from custom checks/presets
- new round of readme improvements to compete `dataproofer`


roll:
@pwalsh 
I still need to integrate tests for it from different branch and improve documentation but it will be great if you will be able to take a high-level look on new checks - https://github.com/frictionlessdata/goodtables-py/tree/implement-contrib-checks#validation-against-advanced-checks (or advise someone to take a look - Brook, Adam, Dan? - if you don't have time on this week)

roll:
@akariv 
Could you please review? 

There is some noise in the diff because of renaming `columns` to `cells` (#198) and adding comments to existent checks but it's isolated in a few commits. It still passes the same tests so all changes in `goodtables.checks` and `goodtables.presets` could be ignored.

roll:
@georgiana-b 
Please review (Adam is off today)

goodtables/contrib/checks/blacklisted_value.py
```diff
@@ -0,0 +1,49 @@
...
+                'column-number': None,
+            })
+
+        # Check value
+        value = cell.get('value')
+        if value in self.__blacklist:
+            message = 'Value "%s" in column %s for row %s is blacklisted'
+            return errors.append({
+                'code': 'blacklisted-value',
+                'message': message % (value, cell['number'], row_number),
```
roll: This one in on purpose to be consistent with Data Quality Spec error messages using column indexes.

roll:
@georgiana-b 
Thanks! I've made changes based on your comments.

PR #211, opened 2017-09-07
Updated to specs-v1 and minor improvements

### LINKED ISSUE #160: Treat "primaryKey" as a column with unique and required constraints
The "primaryKey"(s) for a table must be unique, but, currently, if I set a column to be a "primaryKey" which breaks the uniqueness constraint, it is not flagged.

Probably worth splitting this [JSON fixture](https://github.com/frictionlessdata/goodtables-py/blob/master/data/unique_field.json) into into two: one with a "primaryKey" and one with only a uniqueness constraint.

### LINKED ISSUE #163: Provide visual separation for values inside error messages
# Overview

We need to add quotes for text representation. It's really simple we just need to understand where we should add quotes. So we have to gather some feedback first.

For example:
```
The value {value} in row {row_number} and column {column_number} does not conform to the pattern constraint of {constraint}
```
It will be rendered with value without separators (hard to understand where value boundaries). We need to use something like `value "{value}" in row`.

# Tasks

- [ ] fix for `pattern-constraint` check
- [ ] add quotes for error message templates across checks

### LINKED ISSUE #174: Rebase on specs-v1 implementations
# Overview

We have to migrate to `tableschema-v1` and `datapackage-v1`

# Tasks

- [x] update dependencies
- [x] use `Schema(strict=False)`  in `table` preset (see https://github.com/frictionlessdata/tableschema-py/issues/113)
- [x] use `resource.tabular` in `datapackage` preset
- [x] ensure we support `header=False` from CSV dialect for datapackages
- [x] ensure all required changes are applied

### LINKED ISSUE #206: Nested preset should support individual list of checks for every item
# Overview

For now checks could be customized per `validation` call:

```python
validate('data.csv', checks=['trancated-value'])
```

Sometimes it's very useful to customize checks per table (item) for one validation (see `goodtables.io` case):

```python
validate([
    {'source': 'data1.csv', checks=['check1']},
    {'source': 'data2.csv', checks=['check2']},
])

### LINKED ISSUE #208: Table limit warnings include table source, and are potentially very long
[Table error and row limit warnings](https://github.com/frictionlessdata/goodtables-py/blob/c014ebc33b5d56e2ba2877327d75433bd4981f3b/goodtables/inspector.py#L204-L212) output the table `source` as part of the message. This message can end up very long, with the actually warning message at the end. 

[DPP](https://github.com/frictionlessdata/datapackage-pipelines/) logging has a limit of 64k per log message, so often the error isn't logged.

I'm not sure if the source should be omitted, or an option added to omit it? Or some other way to handle the issue.

### LINKED ISSUE #209: Rebase on Data Quality Spec v1
# Overview

Data Quality Spec is about to be released - https://github.com/frictionlessdata/data-quality-spec/issues/15. We should update our spec.json to v1 and take care of `uncastable-value` to `type-or-format-error` error code renaming.

# Tasks

- [ ] release Data Quality Spec
- [ ] updated `goodtables`

### PR DESCRIPTION
**Pull request by roll**
- fixes #163 
- fixes #209 
- fixes #208 
- fixes #160 
- fixes #206 
- fixes #174

### DIFF
--- a/goodtables/spec.json
+++ b/goodtables/spec.json
@@ -1,5 +1,5 @@
 {
-    "version": "1.0.0-alpha3",
+    "version": "1.0.0",
     "errors": {
 
 
@@ -135,13 +135,13 @@
           "message": "There is a missing header in column {column_number}",
           "description": "Based on the schema there should be a header that is missing in the first row of the data source.\n\n How it could be resolved:\n - Add the missing column to the data source or remove the extra field from the schema\n - If this error should be ignored disable `missing-header` check in {validator}."
         },
-        "non-castable-value": {
-          "name": "Non-Castable Value",
+        "type-or-format-error": {
+          "name": "Type or Format Error",
           "type": "schema",
           "context": "body",
           "weight": 9,
-          "message": "Row {row_number} has non castable value {value} in column {column_number} (type: {field_type}, format: {field_format})",
-          "description": "The value can't be cast based on the schema type and format for this field.\n\n How it could be resolved:\n - If this value is not correct, update the value.\n - If this error should be ignored disable `non-castable-value` check in {validator}. In this case all schema checks for row values will be ignored."
+          "message": "The value {value} in row {row_number} and column {column_number} is not type {field_type} and format {field_format}",
+          "description": "The value does not match the schema type and format for this field.\n\n How it could be resolved:\n - If this value is not correct, update the value.\n - If this value is correct, adjust the type and/or format.\n - To ignore the error, disable the `type-or-format-error` check in {validator}. In this case all schema checks for row values will be ignored."
         },
         "required-constraint": {
           "name": "Required Constraint",
--- a/README.md
+++ b/README.md
@@ -256,7 +256,7 @@ The first row of the data source contains header that doesn't exist in the schem
 
 Based on the schema there should be a header that is missing in the first row of the data source.
 
-#### non-castable-value
+#### type-or-format-error
 
 The value can't be cast based on the schema type and format for this field.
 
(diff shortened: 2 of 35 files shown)

### CODE REVIEW

**Comment by roll** [comment]
@brew 
Could you please review? 
Changes described in commits (also diffs could be easier to check commit by commit).

File: goodtables/spec.json [hunk 1]
```diff
@@ -135,13 +135,13 @@
...
-        "non-castable-value": {
-          "name": "Non-Castable Value",
+        "type-or-format-error": {
+          "name": "Type or Format Error",
           "type": "schema",
           "context": "body",
           "weight": 9,
-          "message": "Row {row_number} has non castable value {value} in column {column_number} (type: {field_type}, format: {field_format})",
-          "description": "The value can't be cast based on the schema type and format for this field.\n\n How it could be resolved:\n - If this value is not correct, update the value.\n - If this error should be ignored disable `non-castable-value` check in {validator}. In this case all schema checks for row values will be ignored."
+          "message": "The value {value} in row {row_number} and column {column_number} is not type {field_type} and format {field_format}",
```
brew: The error type is 'type __or__ format', but the messages are 'type __and__ format'. Will that always be correct?

roll: It's been reviewed on data quality spec level so I suppose it's correct. It's harder for non-native speaker to say but my understanding that:
- value must be this type and format BUT
- value has wrong type or format

roll: Anyway we just use spec from the upstream so it will be addressed on the spec level if needed.

roll:
Thanks @brew!

PR #223, opened 2017-10-05
Fix off-by-one errors in extra_header check

bblough:
Comparison misses the edge case where the extra column has data in it

Resulting assignment does not account for the fact that the column
numbers are 1-indexed while python lists are 0-indexed.


roll:
It's released as `goodtables@1.4.5`. Thanks for your help.

PR #225, opened 2017-10-05
Fix extra_header check to handle empty data strings

bblough:
schema.infer expects a list of lists (rows), per the example at
https://github.com/frictionlessdata/tableschema-py#schema.  However,
extra_header passes the column data as a list of values.  This causes
schema.infer to throw an exception when the data contains one or more
empty strings.


roll:
It's released as `goodtables@1.5.1`

PR #236, opened 2017-11-28
make README.md more accessible

Bonifacio2:
Replacing the report image with actual text makes the README.md accessible to people using screen readers. Since the image doesn't contain more information than plain text this is an easy move. Replacing the other image (the one with a table) without losing something would be more challenging.


vitorbaptista:
@Bonifacio2 Thanks for the PR, this is a good improvement. Unfortunately, I did the same change on https://github.com/frictionlessdata/goodtables-py/pull/235. As that PR changes a lot of the README structure, accepting yours would create merge conflicts. Because of this, I'm closing this PR without merging it.

Thanks again for your contribution. Please feel free to open other PRs or issues :beers:

PR #240, opened 2017-12-15
Refactor error and cells and add report validation JSON Schema

vitorbaptista:
Check the commit messages for the reasoning behind the changes.


roll:
@vitorbaptista
Hey man.. 

I've checked with @pwalsh that I still preserve the role of the Tech Lead of the Frictionless Data libs and tooling. And I hate to say it but my decision is - we don't do this changes now. 95% of this PR is full `goodtables` checks protocol rewrite so I think it's better to preserve this PR for the future and address jsonschema validation (which is great idea) in a new one.

This software stack is far away from being in its infancy allowing us to do any changes we want (esp. related to public interfaces). All core libs here are now in the stable release phase. And it's part of this big software chain, we have a lot of users and services depending on our API (including the `checks` API). We're have an active community around it.

This PR automatically means releasing `goodtables-v2` (according to semver). Based on how Python versioning works it's very possible that this changes still will break our users software. Such an epic release I'm pretty sure it's not what we're looking for as a part of **writing documentation for ODI project** two days before the Cristmass. When there will not be anyone to fix any possible newly introduced bug etc. 

In general I think it's kinda obvious that on this stage of the project (after v1 release) no one including me (author) shouldn't consider to introduce such a big changes without International discussion with the project contributors, users and community in dedicated issue containing a comprehensive design of the proposed changes and motivation.

So if there is still an intention to propose this changes as a part of upcoming `goodtables-v2` work we need to create an issue for the discussion. If quickly my opinion is that I doubt this changes are possible because I have eliminated this design on initial stage as non-working for big data - probably we can't create a `Cell` object for EVERY cell. It's an intention by design of every lib in the stack `tabulator/tableschema/etc` - to use native structures for rows and cells for performance. Even if we can use this approach here we need to do a proper profiling and analysis first as part of the discussion.

vitorbaptista:
@roll 

I agree with you. I should have created an issue and discussed these changes before. I'll do it for any future big refactorings.

On this PR, I changed the code because of the performance issues you raised. The reason for these changes in the first place were to keep the logic of creating a cell in a single place, so instead of creating an `Cell` class, I created a `create_cell()` function that returns a `dict`. The overhead difference is just the function call, which I haven't tested but should be fine.

I kept `Error` as a class because it has some internal logic on generating the error message, and as it isn't created that often, the class overhead should be fine.

With these changes, adding a new custom check becomes:

```python
# Before
@check('custom-check', type='custom', context='body')
def custom_check(errors, cells, row_number):
    for cell in cells:
        errors.append({
            'code': 'custom-error',
            'message': 'Custom error',
            'row-number': row_number,
            'column-number': cell['number'],
        })
        cells.remove(cell)

# After
@check('custom-check', type='custom', context='body')
def custom_check(cells):
    errors = []
    for cell in cells:
        message = 'Custom error on column {column_number} and row {row_number}'
        error = Error(
            'custom-error',
            cell,
            message
        )
        errors.append(error)
        cells.remove(cell)
    return errors
```

The `row_number` is part of the `cell` dict, so it can be removed from the arguments. Also, instead of modifying the `errors` argument to output the errors, we use the return values. I still don't like to be modifying the `cells` argument inside the function to avoid new checks to be run, but this PR is big enough as it is. However, with the `Error` refactoring, I can imagine we removing the cells that have an error attached. Something like:

```python
for check_func in check_funcs:
    errors += (check_func(cells) or [])
    for error in errors:
        if error.cell:
            cells.remove(error.cell)
```

This will make the check functions idempotent. I'll create another issue so we can talk about this.

roll:
```
# After
@check('custom-check', type='custom', context='body')
def custom_check(cells):
```
@vitorbaptista 
There is a reason why `errors` is part of checks API - it's because some checks could require previous errors information.

goodtables/cells.py
```diff
@@ -0,0 +1,73 @@
...
+        'row-number': row_number,
+    }
+    return _remove_none_values(cell)
+
+
+def is_complete(cell):
+    return None not in [
+        cell.get('header'),
+        cell.get('field'),
+        cell.get('value'),
```
roll: That's not how the iteration protocol works - please read https://github.com/frictionlessdata/goodtables-py/issues/247

roll:
@vitorbaptista 
I've bootstrapped the discussion threads with some initial analysis here:
- https://github.com/frictionlessdata/goodtables-py/issues/246
- https://github.com/frictionlessdata/goodtables-py/issues/247

My advice is to start extracting from this gigantic PR some mergeable at the moment parts like JSON Schema addition and go thru it step by step.

goodtables/cells.py
vitorbaptista: You mean that this doesn't handle the case where `cell['header']` exists but is `None`? I can change this to `'header' in cell`, but then the following cell will be considered "complete":

```python
cell = {
  'header': None,
  'field': None,
  'value': None,
  'column-number': None
}
```

Which seems wrong, but I might be missing something.

vitorbaptista:
> There is a reason why `errors` is part of checks API - it's because some checks could require previous errors information.

Could you give me an example of such check?

vitorbaptista:
@roll I wrote some specific comments in the issues you opened, but TBH I don't understand what you're arguing. The epic issues you opened do a good job explaining the constraints we need to consider, but I don't see how this PR fails these constraints. The only argument I got was that this is a breaking change, which is true, but that's how software evolves. I see this PR as an improvement because it encapsulates the logic of Error and cells creation in a single place, making it easier to change them if needed.

Do you have any specific suggestions for improvement?

roll:
@vitorbaptista 
I think this PR contains really great ideas:
- JSON Schema validation
- renaming `number` to `column-number` (initially it was column instead of cell)
- moving `row-number` inside the `cell` dictionary

roll:
@vitorbaptista 
I've added the above comment with changes I agree:
- JSON Schema validation
- renaming number to column-number (initially, it was `column` instead of `cell`)
- moving row-number inside the cell dictionary

There is some refactoring which seems good to me. Though I'm not a supporter of mixing actual changes with refactoring on the same PR because it's always a risk to introduce new issues.

Also, there are internal changes I think is not good in my opinion e.g. creating dependencies for constraint checks. But it's not important in comparison to breaking changes. Though I've re-checked it and just realized that there is new function creation with closure for every row for every constraint check - should be profiled for big files.

Regarding other stuff:
- as I've mentioned there is incorrect `cell` key existent checks across the whole PR (null checks instead)
- there is `is_complete` thing that I don't see also as a correct thing. Because every check relies on some subset of available cell keys (see https://github.com/frictionlessdata/goodtables-py/issues/247) not sure what does it mean to be complete or not. E.g. if a user hasn't provided a schema a `cell` will not be having `cell['field']` still being complete having `header` and `value`.

But yes, the main problem is removed `errors` and added `Error` class that:
- removes the direct errors access from checks (not flexible enough)
- create errors list for every row for every check (have to be profiled for big files)

Having `errors` access is a design decision to do not guess. So I just can't predict what kind of checks will be created. Just pretty sure that direct access will be needed at some point (esp. having `goodtables.io` now which consumes goodtables reports without any ability to edit it on user-level). 

Because this interface is low-level and mostly for OKI/partners (see ckan extensions) this `just in case` looks good for me unlike to other situations.

vitorbaptista:
I just pushed a few more commits where I:

* Optimized the constraint checks function creation, now the function is created only once when that file is loaded, not on every call (nice catch).
* Reviewed all checks for cells attributes, modifying between `cell['foo'] is None` or `'foo' in cell` as needed by each check (my review of all checks is below)
* Removed the `cells.is_complete()` method, each check will verify whatever it needs

## Review of `'foo' in cell` vs `cell.get('foo') is not None`

Regarding `'foo' in cell` vs `cell.get('foo') is not None`. Let's consider a cell as:

```python
cell = {
  'header': None,
  'value': None,
  'number': None,
  'field': None,
}
```

* [duplicate_header][dh]
  It'll complain that there are duplicate `None` headers. Seems to be what we want.
* [extra_header][eh]
  It will skip the cell, which doesn't seem correct. All cells must have non-null `'field'` coming from the schema, otherwise an error must be raised (or the field inferred).
* [extra_value][ev]
  It will skip the cell, which similarly to `extra_header`, also doesn't seem correct. I don't remember why, but this PR changes this check quite a bit.
* [missing_header][mh]
* [missing_value][mv]
  These are similar to the `extra_header` and `missing_header`. For instance, for our example `cell` with `cell['value'] == None`, should we raise a `missing-value` error? It seems like we should.
* [non_matching_header][nmh]
  An exception will be thrown because we're trying to get `cell['field'].name`. If `cell['field']` wasn't `None` but `header` and `number` were, we should raise an error because `cell['header'] != cell['field'].name`. Added a test and fixed my PR.
* [pattern_constraint][pc]
* [required_constraint][rc]
* [type_or_format_error][tfe]
* [unique_constraint][uc]
  It will throw an exception because we're trying to get `cell['field'].descriptor`, so checking that `cell['field'] is not None` is the correct behaviour.
* [custom_constraint][cc]
  Initially I thought `cell['header']` could be `None` (but not `cell['value']`), because it could be a blank header. However, I realised that there's no way to reference a blank header in a custom constraint, so both header and value needs to be defined.
* [create_check_constraint][ccc]
  This isn't on master, but as there are many checks that are basically copy and pasted, it's easier to talk about them all using the `create_check_constraint`. They used the `set(cell).issuperset(['number', 'header', 'field', 'value'])` (e.g. https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/enumerable_constraint.py)
  With our example `cell`, this would raise an error when we tried to call `cell['field'].test_value()`. That seems to be the only requirement, though: `cell['field']` can not be `None`.


[dh]: https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/duplicate_header.py
[eh]: https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/extra_header.py
[ev]: https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/extra_value.py
[mh]: https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/missing_header.py
[mv]: https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/missing_value.py
[pc]: https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/pattern_constraint.py
[rc]: https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/required_constraint.py
[tfe]: https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/type_or_format_error.py
[uc]: https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/unique_constraint.py
[cc]: https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/contrib/checks/custom_constraint.py
[ccc]: https://github.com/frictionlessdata/goodtables-py/blob/092bc917154301ca9b12185c75af077bb7d2f576/goodtables/checks/constraints_checks.py
[nmh]: https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/non_matching_header.py

roll:
@vitorbaptista 
Thanks. I'm on it. Wasn't able to finish today. Going to finish tomorrow early morning UTC.

roll:
@vitorbaptista 
Thanks. I'm on it. Wasn't able to finish today. Going to finish tomorrow early morning UTC.

upd.
accidentally closed/opened

roll:
@vitorbaptista 
I've tried to do my best reviewing this PR jumping many times here from other work for 1-2h. TBH I've mostly failed because related to the complex things my brain starts to properly work only after a few days being in the context. So I still look at this changes and don't understand 50% of this. It seems almost every moving part of goodtables has been changed in this PR. At least I hope I was able to help to find some problems and improve design decisions sharing goodtables architecture knowledge. As you know I'm a big supporter of design consistency based on high-level vision (on this one I and OKI have slightly differing views). It's your project going forward so if you're sure It works and properly profiled please go for it and merge these changes. Regarding me at least I feel much better after things like `is_complete` has been removed. Still, disagree with overall direction and the way of introducing such big and critical changes but I'm sure you know what you're doing. 

PS.
Regarding to `'foo' in cell vs cell.get('foo') is not None`. Probably I wasn't fully correct. I mean that this is the same as `==` and `===` in JavaScript or race conditions in threading. In many cases, the result will be the same. My point was that `'foo' in cell vs cell.get('foo') is not None` are just different checks. So an idea was to use consistently correct one because erasing a border between these checks could lead to confusing bugs. Unfortunately, due to the performance reasons, we can't use here better object-oriented interface.

Issue #233, opened 2017-11-23
Review the CLI API


vitorbaptista:
This is our current CLI help:

```
Usage: goodtables [OPTIONS] SOURCE

  https://github.com/frictionlessdata/goodtables-py#cli

Options:
  --preset TEXT
  --schema TEXT
  --checks <LAMBDA>
  --skip-checks <LAMBDA>
  --infer-schema
  --infer-fields
  --order-fields
  --error-limit INTEGER
  --table-limit INTEGER
  --row-limit INTEGER
  --json
  --version               Show the version and exit.
  --help                  Show this message and exit.
```

If you don't know what these options mean, it's not helpful

## Tasks

- [x] Validate that the parameters (e.g. `row-limit` should be greater than zero 0)
- [x] Allow reading from STDIN
- [x] Allow glob paths (https://github.com/frictionlessdata/goodtables-py/issues/229)
- [x] Review defaults (maybe we could remove the 1000 rows limit?)
- [x] Explain what goodtables does, and each of its options
- [x] Add `--quiet` option
- [x] Support short and long options (e.g. `-v` and `--verbose`)
- [ ] Consider not outputting anything when the data is valid (as is costumary in UNIX)

##  Resources

* https://zapier.com/engineering/how-to-cli/
* http://eng.localytics.com/exploring-cli-best-practices/
* http://docopt.org/
* https://softwareengineering.stackexchange.com/questions/307467/what-are-good-habits-for-designing-command-line-arguments
* http://www.catb.org/esr/writings/taoup/html/ch10s05.html
* https://medium.com/re-form/ux-for-wizards-e6ffaad577b4

vitorbaptista:
Fixed

PR #241, opened 2017-12-15
Improve the CLI

vitorbaptista:
- connects #233

---

Please check the commit messages for the reasoning.


vitorbaptista:
I agree the CLI options should be the same as the other options. I've done it on the CLI first so we can try and discuss, and then, if we agree, change on the other parts.

Regarding `ignore-order`, the intent is renaming it for what the user wants to do, not the internal implementation. What she wants is to ignore the fields order. We can rename it to `ignore-fields-order` to make that clearer, WDYT?

P.S.: I think we should release a new major version, as there's no concept of "provisional API" on semantic versioning when the version is >= 1.x. However, this release can wait until other changes are merged.

vitorbaptista:
I went back to using `order-fields`, so there are no naming changes. This only improves the documentation, allows validating multiple files and some other minor CLI usability improvements (output to file, support files from STDIN, documentation, etc.), and added the `goodtables init` command to generate an initial datapackage.

roll:
> I went back to using order-fields, so there are no naming changes.

Great. Without name changes, we could be sure that everything is OK with goodtables-js, goodtables.io/API, ckanext-validation and other clients.

PR #252, opened 2018-02-23
[#251] Allow non-required primary keys

### LINKED ISSUE #251: Required Check is Overloaded with Primary Key Check
I'm working with a messy public dataset where some of the fields that I'd consider part of the "primary key" for that row are null.

Unfortunately, even if I specify that this field's constraints as  `required: False`, because they are part of the primary key, they will throw an error because of the line below.

https://github.com/frictionlessdata/goodtables-py/blob/75c7f019b11ba4c6045bbb8338ecc60590a887d9/goodtables/checks/required_constraint.py#L24

Two possible ideas come to mind to resolve this:

1. Don't overload this check and remove the primary key test
2. Let a "required = False" override the primary key check

vitorbaptista:
If a field is a `primaryKey`, we validate that it exists, the same way as if it
had `{'constraints': {'required': True}}`. However, there can be some cases
where a `primaryKey` field can be null. Pretty uncommon, but messy data...

This commit allows removing the check by explicitly setting the `required` to
False, so a field like:

```
"fields": [
  {
    "name": "id",
    "constraints": {
      "required": False
    }
  }
],
"primaryKey": "id"
```

Wouldn't be required. In other words, the `required` constraint overloads the
`primaryKey` constraint.

Fixes #251

### DIFF
--- a/goodtables/checks/required_constraint.py
+++ b/goodtables/checks/required_constraint.py
@@ -26,7 +26,7 @@ def required_constraint(cells):
 
         # Check constraint
         valid = field.test_value(value, constraints=['required'])
-        if field.descriptor.get('primaryKey'):
+        if field.descriptor.get('primaryKey') and field.constraints.get('required', True):
             valid = valid and field.cast_value(value) is not None
 
         # Skip if valid
--- a/tests/checks/test_required_constraint.py
+++ b/tests/checks/test_required_constraint.py
@@ -4,13 +4,52 @@ from __future__ import print_function
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+from tableschema import Field
 from goodtables.checks.required_constraint import required_constraint
+import goodtables.cells
 
-
-# Check
 
 def test_check_required_constraint(log):
-    cells = []
-    errors = required_constraint(cells)
-    assert log(errors) == []
-    assert len(cells) == 0
+    cell = goodtables.cells.create_cell(
+        'id',
+        '',
+        field=Field({
+            'name': 'id',
+            'constraints': {
+                'required': True,
+            }
+        })
+    )
+    errors = required_constraint([cell])
+    assert len(errors) == 1
+    assert errors[0].code == 'required-constraint'
+
+
+def test_primary_key_fields_are_required_by_default(log):
+    cell = goodtables.cells.create_cell(
+        'id',
+        '',
+        field=Field({
+            'name': 'id',
+            'primaryKey': True,
+        })
+    )
+    errors = required_constraint([cell])
+    assert len(errors) == 1
+    assert errors[0].code == 'required-constraint'
+
+
+def test_primary_keys_required_constraint_can_be_overloaded(log):
+    cell = goodtables.cells.create_cell(
+        'id',
+        '',
+        field=Field({
+            'name': 'id',
+            'constraints': {
+                'required': False,
+            },
+            'primaryKey': True,
+        })
+    )
+    errors = required_constraint([cell])
+    assert not errors


Stephen-Gates:
I'm not convinced this is a good idea. A primary key test should test for present and unique values. If they're not present or unique the test should fail.

So this schema will give a different result to the example above?

"fields": [
  {
    "name": "id"
  }
],
"primaryKey": "id"

Seems like this approach could result in errors if a UI has a switch to toggle `required` on/off and does this by setting `true`/`false` in the `constraint` rather than setting it to `true` or not setting the `constraint`.

roll:
> However, there can be some cases where a primaryKey field can be null.

If we follow the SQL meaning of the primary key term it just can't be `not a value`. That's my understanding. 

The other question that the specs say a lot about how to writer/format primary keys for Table Schema but almost nothing about what having a primary key means for the data. Probably the narrative could be extended at the specs level also clarifying the `null` situation. (cc @rufuspollock)

vitorbaptista:
This is tricky. According to the SQL standard, a primary key column is always NOT NULL, so this PR deviates from the standard. If we allow null primary keys, what implications will it have on foreign keys? We might end up opening a can of worms.

@hydrosquall could you explain more why you'd need this? It sounds like the correct answer is either 1) fixing the source data, or; 2) not using nullable fields as part of the primary key. A third option would be allowing the primarykey checks to be disabled, either by extracting its checks into a `primarykey-constraint`, or adding some option to `required-constraint` and `unique-constraint` that disables primaryKey checks, but I'm not sure.

Stephen-Gates:
I wonder if using the [unique constraint](https://frictionlessdata.io/specs/table-schema/#constraints) would address your requirement @hydrosquall?

hydrosquall:
@roll / @Stephen-Gates : thanks for the link. The issue here is I don't want to enforce uniqueness on an individual field, I'm OK with duplicates on a single field. What I need is for a "composite key" derived from multiple fields to be unique, without actually putting a new field into my data.

Stephen-Gates:
A [`primaryKey`](https://frictionlessdata.io/specs/table-schema/#primary-key) can be made up of more than one field as shown in the example in the spec and repeated below:

```
"schema": {
  "fields": [
    {
      "name": "a"
    },
    {
      "name": "b"
    },
    {
      "name": "c"
    },
    ...
  ],
  "primaryKey": ["a", "c"]
 }
```

and so I think you want this data to be valid with the schema above: 

a | b | c
-|-|-
1 | | a
1 | | 
2 | a | a
  | | | a

Neither field a or c are individually `required` or `unique`.
Together as a key, a and c, must be `required` and `unique`.

Adding this row to the data would be invalid

a | b | c
-|-|-
&nbsp;  | a | 

Is this want you want @hydrosquall ?

hydrosquall:
@Stephen-Gates : thanks for the clean example.

The way this issue request came up is because this is how I was using primary key: specifying something like `a` and `c` make up the primary key. However, I don't want the requirement that `a` and `c` are `not null`.

In the example above, I would accept the example row with the lone `a` because there aren't any existing rows where both `a` and `c` are null in the receiving table. 

I would like to reject the following new rows though because their primary key already exists.

a|b|c
--- | --- | ---
1|2|a
|1|a

vitorbaptista:
So it seems what you really want is a composite unique key, and partial indexes. The specification doesn't support this yet, but I created an issue to discuss this on https://github.com/frictionlessdata/specs/issues/593.

This isn't a trivial change, so it might take some time. Meanwhile, I see two options:

* Create a custom check
* Overwrite the `required_constraint` with the code in this PR

For the custom check, you'd need to configure each check validation with the unique field names, similar to https://github.com/frictionlessdata/goodtables-py/#custom-constraint. This is because, as far as I know, a check isn't able to read the table schema, so you can't add a custom attribute like `_uniqueKeys: ['id_a', 'id_b']` (@roll am I wrong?)

Overwriting the `required_constraint` seems to be the easiest way :+1:

I'm closing this PR. We can continue the discussion on https://github.com/frictionlessdata/specs/issues/593.

PR #256, opened 2018-03-03
[WIP] Use separate processes for parallelizing table checks

### LINKED ISSUE #255: Use separate processes instead of threads for paralellizing tasks
I've recently been setting up a nightly task that uses goodtables to validate a data package with 15 tables including one with about 1.5M lines and nine others with over 20k. I'm glad to see goodtables churning through them all in parallel, but would love it not to be constrained to a single CPU core.

reidab:
Decided to take a quick stab at #255. I haven't worked with Python much, so critique here is more than welcome.

This allows validation of multi-table data packages to span across multiple CPU cores! 🎉 

This is implemented using the `Manager` and `Process` classes directly, because using `multiprocessing.Pool` requires that everything involved can pickle nicely into a `multiprocessing.Queue`. Sadly jsonschema's dynamically-generated `Validator` classes don't seem to play well with pickling. I tried switching to `pathos.multiprocessing`, which uses `dill`, but this also failed with the jsonschema `Validator`.

Fixes #255

### DIFF
--- a/goodtables/inspector.py
+++ b/goodtables/inspector.py
@@ -11,7 +11,7 @@ import tabulator
 import tableschema
 from . import cells
 from copy import copy
-from multiprocessing.pool import ThreadPool
+from multiprocessing import Process, Manager
 from .registry import registry
 from .error import Error
 from . import exceptions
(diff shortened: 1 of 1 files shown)


roll:
@reidab 
A table schema could be serialized as `schema.descriptor` dict and restored as `Schema(descriptor)` inside a process. 

I've added a few comments to #255. Curious what benchmarks show after this change.

Thanks for the help!

reidab:
I probably won't have time to dive back into the serialization stuff until next weekend, but if anyone else wants to look at it I think things were breaking down when trying to pass `table` to `_inspect_table` via a `Queue`.

On the other hand, using a pool here doesn't really gain much more than a slight increase in clarity, so it may be good to go as-is.

One optimization would be to only spawn processes if there's more than one table, since for a single table we can do all the checking in the main thread.

roll:
Thanks @reidab I'll finish this PR as a part of https://github.com/frictionlessdata/goodtables-py/issues/341

PR #266, opened 2018-04-12
Fix Markdown description on PyPI

serahrono:


goodtables/VERSION
```diff
@@ -1 +1 @@
-2.0.1
+2.0.2
```
roll: @serahrono 
We would better to have the version change as the separate commit. Please take a look at the commit - https://github.com/frictionlessdata/goodtables-py/commit/3f0d442fb0931e8bf57f4a285835ce049e7d1803.

It allows having the commit history with an explicit release (version) commits  - https://github.com/frictionlessdata/goodtables-py/commits/master

PR #278, opened 2018-09-14
Dynamic config

pierredittgen:
# Dynamic config

Make goodtables-py a bit more flexible to allow:
- use of an external (e.g. localized) spec.json
- use of user-defined custom-checks defined out of "goodtables-py" project

## Using cli
No change. Use cli.py as usual

### Init
Before using library, first call:
```python
goodtables.init()
```
or if you want to use a customized spec.json file:
```python
goodtables.init(spec_path='/path/to/my/alternate_spec.json')
```

### Using external custom-check
Define your custom check script file inside your client project, e.g in **/path/to/my_project/checks/foo_value.py**:
```python
from goodtables.registry import check
from goodtables.error import Error

@check('foo-value', type='custom', context='body')
class FooValue(object):
    """ Checks if a cell value equals 'foo' """

....
```
Then load class from the client project code:
```python
goodtables.load_check_from_module('/path/to/my_project/checks/foo_value.py')
```

and pass its configuration to Inspector:
```python
checks = ['structure', 'schema', {'foo-value':{'column':'bar'}]
inspector = goodtables.Inspector(checks=checks)
```

Tip: use `goodtables.load_checks_from_dir` if you need to load all custom check classes from a folder.

## Notes

This pull-request was initiated to ease the authoring of custom checks in the [Validata](https://dev.validata.fr/) project, which is targeted for a French audience, and needs both a [localized `spec.json` file](https://github.com/Jailbreak-Paris/goodtables-py/blob/1d8b68c7f33630ff84101ef9e1381c769e9cb5ea/goodtables/spec.json) and French-specific custom checks like [`FrenchSiretValue`](https://github.com/Jailbreak-Paris/goodtables-py/blob/d82a165436b25bdb71847d48320f5d3a3ab9343d/goodtables/contrib/checks/french_siret_value.py).


roll:
@pierredittgen 
Hi, thanks!

But AFAIK we can do now:

```
import you_custom_checks_module

checks = ['structure', 'schema', {'foo-value':{'column':'bar'}]
inspector = goodtables.Inspector(checks=checks)
```

because `@check/preset` uses the global registry.

So this actually does the same out-of-box as:

```
goodtables.load_check_from_module('/path/to/my_project/checks/foo_value.py')
```

pierredittgen:
Hi @roll,

Thanks for your answer, you're totally right on custom_check import.

I've opened a new  [issue](https://github.com/frictionlessdata/goodtables-py/issues/279) about using localized spec.json with goodtables.

PR #280, opened 2018-09-18
Add message parameter values to error info dict

pierredittgen:
Add message parameter values to error information dict (key `message-data`). This allows alternative usage of error description.

Before:
```json
{
    "code": "type-or-format-error",
    "column-number": 8,
    "message": "The value \"5500,20\" in row 4 and column 8 is not type \"number\" and format \"default\"",
    "row-number": 4
},

```
After:
```json
{
    "code": "type-or-format-error",
    "column-number": 8,
    "message": "The value \"5500,20\" in row 4 and column 8 is not type \"number\" and format \"default\"",
    "message-data": {
        "field_format": "default",
        "field_type": "number",
        "value": "5500,20"
    },
    "row-number": 4
},

```

### DIFF
--- a/goodtables/error.py
+++ b/goodtables/error.py
@@ -106,4 +106,5 @@ class Error(object):
             'row-number': self.row_number,
             'column-number': self.column_number,
             'message': self.message,
+            'message-data': {k: v.strip('"') for k, v in self._message_substitutions.items()},
         }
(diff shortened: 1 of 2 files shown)


roll:
@pierredittgen 
Great idea!

Could you please fix the implementation to pass the tests and rebase the branch?

PR #286, opened 2018-10-01
Fix duplicate_header check

cbenz:
Hi,

I noticed that, in the `duplicate_headers` check, [the for-loop variable `cell`](https://github.com/frictionlessdata/goodtables-py/blob/7778d1e3630f4ab95dfb76d13106c9733d4ec4c0/goodtables/checks/duplicate_header.py#L18) is used after the loop ([here](https://github.com/frictionlessdata/goodtables-py/blob/7778d1e3630f4ab95dfb76d13106c9733d4ec4c0/goodtables/checks/duplicate_header.py#L40)).

The consequence is that the cell mentioned by the `goodtables.Error` is always the last one. The only existing test was exactly testing this situation: when I updated it, it failed.

I modified the check to fix the reference of the duplicated column, and to designate the first occurrence of the column in the message.

Also: should this check return as many errors as duplicate headers, or just one mentioning all the duplicates? I chose the first solution with many errors.

What do you think about it? I'm not 100% sure about my implementation since I modified an existing algorithm, and did not start from scratch.

Thanks


roll:
@cbenz 
Thanks! I will take a look. Interesting if the data quality spec provides some recommendations on it.

roll:
Sorry, I'm not able yet to review. I hope to do it soon.

PR #287, opened 2018-10-08
Pass raw data to error message_substitutions

cbenz:
I'm building a UI using goodtables-py as a dependency, and I need to display raw data which is currently prisoner of the error message string.

For example, I would like to display a `<ul>` of all possible values of an enum constraint.

The previous pull request #280 allowed to export `messages_substitutions` as `message-data` in error dict.
But many checks format values (using `", ".join` for example), making those data difficult to reuse.

This pull request modifies those checks to expose data and a formatted version of the data (for example exposing both keys `column_numbers` and `column_numbers_str`) and removes the stripping of quotes which become unnecessary.

I also modified `spec.json` to add quotes in the message string.

I noticed that this approach is already adopted in custom checks (example in [`blacklisted_value.py`](https://github.com/frictionlessdata/goodtables-py/blob/7778d1e3630f4ab95dfb76d13106c9733d4ec4c0/goodtables/contrib/checks/blacklisted_value.py#L45-L51)).

### DIFF
--- a/goodtables/checks/constraints_checks.py
+++ b/goodtables/checks/constraints_checks.py
@@ -26,8 +26,8 @@ def create_check_constraint(check, constraint):
             # Add error
             if not valid:
                 message_substitutions = {
-                    'value': '"{}"'.format(value),
-                    'constraint': '"{}"'.format(field.constraints[constraint]),
+                    'value': value,
+                    'constraint': field.constraints[constraint],
                 }
 
                 error = Error(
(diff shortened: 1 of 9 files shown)


roll:
@cbenz 
Thanks! Sorry for the VERY late reply..

The problem with changing `spec.json` is that we have to follow https://github.com/frictionlessdata/data-quality-spec. It's used e.g. but UI like goodtables.io's fronted etc. So we need to propose changes to the spec first. Probably it's better to try to work with current one at the moment.

johanricher:
Thanks for the answer @roll. Would the PR be accepted if there's a proposition to modify `spec.json` in the `data-quality-spec` repo first?

roll:
@johanricher 
I'm not sure. It's just established `v1` and it will be breaking changes. 

I'm not sure but maybe it's possible to provide user's `spec` to the `goodtables.validate` function? Customization on this level could be much more flexible.

PR #292, opened 2019-04-10
fix: Make unique and primary constraint errors more helpful

### LINKED ISSUE #291: Unique constraiant error message includes too many row numbers
When validating data with a unique constraint, the error message will include all row numbers between the last valid row and the row that violates the constraint. 

See https://gist.github.com/jdaviscooke/9aa1f605a8fdd946c23bf023848d0843 for an example of how I am running into this error. 

I would expect to see that rows 4 and 10 violate the unique constraint, but instead I see
`
{
          "code": "unique-constraint",
          "row-number": 10,
          "column-number": 1,
          "message": "Rows 2, 3, 4, 5, 6, 7, 8, 9, 10 has unique constraint violation in column 1",
          "message-data": {
            "row_numbers": "2, 3, 4, 5, 6, 7, 8, 9, 10"
          }
        }
`
Happy to provide any additional detail that you need.

jdaviscooke:
Fixes #291 

Changes error messages on unique or primary key constraints to include only the currently conflicting row, as well as the column headers and the conflicted values.

```json
   {
          "code": "unique-constraint",
          "message": "Row 10 has unique constraint violation in column(s) 'id'. Conflicting value(s) '1'",
          "message-data": {
            "row_numbers": 10,
            "value": "1",
            "type": "unique",
            "header": "id"
          }
        },
        {
          "code": "unique-constraint",
          "message": "Row 10 has primary key constraint violation in column(s) 'id, name'. Conflicting value(s) '1, Paul'",
          "message-data": {
            "row_numbers": 10,
            "value": "1, Paul",
            "type": "primary key",
            "header": "id, name"
          }
    }
```


roll:
@jdaviscooke 
Thanks! The problem with changing `spec.json` is that we have to follow https://github.com/frictionlessdata/data-quality-spec. It's used e.g. but UI like goodtables.io's fronted etc. So we need to propose changes to the spec first. Probably it's better to try to work with current one at the moment.

jdaviscooke:
@roll got it, thanks. I'll have another look and see what I can do without changing `spec.json`. Admittedly I think the current spec doesn't provide a good way of surfacing important information about the unique/primary key constraint errors, but I'm happy to stick with it for now and propose a change to the spec separately.

jdaviscooke:
Ok, I've reverted the change to `spec.json` and the related changes to feed that information into `spec.json`. This still cleans up the error messaging to remove the duplicates, so it looks like this now
```json
 "errors": [
        {
          "code": "unique-constraint",
          "row-number": 4,
          "column-number": 1,
          "message": "Rows 4 has unique constraint violation in column 1",
          "message-data": {
            "row_numbers": "4"
          }
        },
        {
          "code": "unique-constraint",
          "row-number": 9,
          "column-number": 1,
          "message": "Rows 9 has unique constraint violation in column 1",
          "message-data": {
            "row_numbers": "9"
          }
        },
        {
          "code": "unique-constraint",
          "row-number": 10,
          "column-number": 1,
          "message": "Rows 10 has unique constraint violation in column 1",
          "message-data": {
            "row_numbers": "10"
          }
        }
      ]
```

roll:
I think we can fix this issue without dropping this additional information completely. For example, adding a message like:

```
"message": "Rows 2, 3, 4, 5 (and 100 more rows) has unique constraint violation in column 1"
```

PR #295, opened 2019-05-09
Allow headers containing None values to appear in report

pierredittgen:
When validating Excel files containing extra columns, header list can look like:
```python
["id", "name", "age", None, None, None]
```
In that case, no header information is returned in report. It's a pity, because part of this header list is usable.


roll:
Thanks! It was a bug

PR #298, opened 2019-05-19
Fix KeyError "field" for reordered columns and don't return  non-matching-header when it's actually a missing-header

delenamalan:
Addresses https://github.com/frictionlessdata/goodtables-py/issues/284 and https://github.com/frictionlessdata/goodtables-py/issues/294


roll:
@delenamalan 
Thanks! I'll review as soon as I'm around

Issue #312, opened 2019-10-09
None of header validators(inbuilt and custom that has context="head") are invoked if there is a blank header


SrividhyaGuru:
# Overview

If the excel has a blank header column in between, none of the validations for headers are triggered.

```
report = validate(file_path, skip_rows=[1, 2, 3, 4], checks=[
        'blank-header', 'missing_header_check', 'duplicate-header'
    ])
```
<img width="731" alt="Screenshot 2019-10-09 at 6 46 07 PM" src="https://user-images.githubusercontent.com/24490498/66484867-70985880-eac5-11e9-92c7-9137f4e43289.png">

Since 5th column is having blank headers, we wanted to see the blank header validator report that but it did not. Also we had **missing_header_check** that is custom built check that has context="head". But **missing_header_check** also did not kick in.

---

Please preserve this line to notify the @frictionlessdata/python team

roll:
@SrividhyaGuru 
Could you please share some file without real data that breaks `goodtables` that way?

SrividhyaGuru:
Schema used:
```{
  "fields": [
    {
      "name": "Column_1",
      "title": "Column_1",
      "type": "string",
      "description": "Sample Column_1",
      "constraints": {
        "required": true
      }
    },
    {
      "name": "Column_2",
      "title": "Column_2",
      "type": "string",
      "description": "Sample Column_2",
      "constraints": {
        "required": true
      }
    },
    {
      "name": "Column_3",
      "title": "Column_3",
      "type": "string",
      "description": "Sample Column_3",
      "constraints": {
        "required": true
      }
    },
    {
      "name": "Column_4",
      "title": "Column_4",
      "type": "string",
      "description": "Sample Column_4",
      "constraints": {
        "required": true
      }
    }
  ]
}
```

The following excel is missing a column **Column_4** that is mentioned in schema. The excel has an extra header **Column_5** which is not mentioned in the schema. And **the excel is having a blank header in column D**

[missing_column4_and_extra_header_column5.xlsx](https://github.com/frictionlessdata/goodtables-py/files/3713928/missing_column4_and_extra_header_column5.xlsx)

Running validate with the above mentioned excel with the above mentioned schema with checks as
`checks=["missing-header", "blank-header", "non-matching-header"]`
yields the following report:

```{'error-count': 0,
 'preset': 'table',
 'table-count': 1,
 'tables': [{'error-count': 0,
             'errors': [],
             'format': 'xlsx',
[...]

PR #314, opened 2019-10-15
Fix header checks getting blocked when one or more blank header is pr…

SrividhyaGuru:
…esent (#312)

* Dont block header checks if there is a None in headers list

* Fix Blank header check to give error when header is not present

* Fix extra value check to report error only when the cell actually has a value in it

* Report missing value error only for columns with headers

# Overview

---

Please preserve this line to notify @roll (maintainer of this repository)


roll:
Thanks! I'll review it in a few days.

goodtables/checks/missing_value.py
```diff
@@ -13,12 +13,16 @@
...
+    missing-value: 	A row has less columns than the header.
+    """
     errors = []
 
     for cell in copy(cells):
 
         # Skip if cell has value
-        if cell.get('value') is not None:
+        # Skip if cell is missing value but belongs to blank header
+        if cell.get('value') or not cell.get('header'):
```
roll: I think it shouldn't be related. A header could be empty `''` in csv by mistake but it still creates a column

SrividhyaGuru: @roll Thanks for reviewing

Missing value check provided the following definition in Readme.
 
Screenshot 2019-10-24 at 10 35 34 PM

If csv had blank header, and if the empty cells under it are reported as missing values, it contradicted the definition of the missing values. 
Thus added the check. 

Please validate if empty cells under a blank header(which itself is not a proper header) can be reported as missing values.

If so, will make necessary changes and push.

roll: @SrividhyaGuru

> Please validate if empty cells under a blank header(which itself is not a proper header) can be reported as missing values.

Yes. This statement is correct. They have to. Sorry for the confusion We need to add a note a readme. The idea that the `headers` row determines a dimension of the table so, for example:

```csv
id,,question2
1,no,yes
2,yes,no
3,,no
```
This table has 3 columns and both the header and row 3 miss values for column 2

SrividhyaGuru: @roll thanks for the clarification. Will update the code soon.

SrividhyaGuru: @roll Have reverted the check for missing value check. Earlier missing value validator checked only for None and not for empty. Now the condition `if cell.get('value')` will cover both None and empty and report as missing values.
The build is breaking because of test_foreign_key.py.
This is due to the below code in **missing_value.py**

       # Add error
        error = Error('missing-value', cell)
        errors.append(error)

        # Remove cell
        cells.remove(cell)

The missing value check method after adding the missing value error, removes the cell itself. The rest of the validators dont get that cell at all. 
Thus foreign key check did not get a cell that had missing value for next_id

Is it necessary to remove the cell from all other validation if it is missing value?

SrividhyaGuru: @roll Could you please confirm on the above so that we could make the necessary changes. If cells could be removed if it has missing values, it would impact the foreign key validator that plays after the missing value validator.

roll: Hi, thanks for your help.

I've been a little bit busy lately. I'm going to wrap my head around it this week or next week.

SrividhyaGuru: Sure, thanks @roll

PR #315, opened 2019-10-28
Return file not found error if file in CLI does not exist

### LINKED ISSUE #265: Goodtables CLI should fail if received paths doesn't exist
For example:

```
$ goodtables foobar                                       
DATASET
=======
{'error-count': 1,
 'preset': 'nested',
 'table-count': 1,
 'time': 0.002,
 'valid': False}

TABLE [1]
=========
{'error-count': 1,
 'headers': [],
 'row-count': 0,
 'source': 'foobar',
 'time': 0.001,
 'valid': False}
---------
[-,-] [format-error] Format "None" is not supported
```

It should have failed with `[-,-] [io-error] [Errno 2] No such file or directory: 'foobar'`, as it already does if the file ends in `.csv`

tsungyong:
# Overview

Fixes #265 

Added a warning to indicate if the file does not exist on the command line.  I opted to do this in the table instead of in the validator, since by that point the Stream has already been created (which is where the `Format "None" is not supported` error is coming from, when it attempts to run open() on that stream).

---


tsungyong:
I've updated it.  I wonder if it would be a bit cleaner if this was implemented in tabulator.Stream though, as the root issue is that `Stream.open('invalid.csv')` is giving a different error from `Stream.open('invalid')`, even though they are both the same error (the local file is not found).

PR #317, opened 2019-11-13
161 support optional fields

JoeStaines:
# Overview

First pass at supporting optional fields via passing a list through `validate`. Does this by not creating the cells for optional fields which are defined in the schema, but the header and value for that field are not present. Not sure if this is the best way forward for optional fields, or if there are some unintended consequences of doing it this way as i'm still unfamiliar with the overall system, so input is welcome.

---

### DIFF
--- a/goodtables/cells.py
+++ b/goodtables/cells.py
@@ -1,13 +1,17 @@
 from six.moves import zip_longest
 
 
-def create_cells(headers, schema_fields, values=None, row_number=None):
+def create_cells(headers, schema_fields, optional_fields, values=None, row_number=None):
     """Create list of cells from headers, fields and values.
 
     Args:
         headers (List[str]): The headers values.
         schema_fields (List[tableschema.field.Field]): The tableschema
             fields.
+        optional_fields (List[str]): List of schema fields which are defined as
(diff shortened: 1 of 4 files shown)


roll:
@JoeStaines 
Thanks!

I'll review at the end of the week or at the start of the next one

roll:
@JoeStaines 
I've taken a look, it seems like a good start but I didn't understand how it integrates with high-level `validation`.

Could you please add some tests showing how the feature works?

goodtables/cells.py
```diff
@@ -1,13 +1,17 @@
...
 
-def create_cells(headers, schema_fields, values=None, row_number=None):
+def create_cells(headers, schema_fields, optional_fields, values=None, row_number=None):
     """Create list of cells from headers, fields and values.
 
     Args:
         headers (List[str]): The headers values.
         schema_fields (List[tableschema.field.Field]): The tableschema
             fields.
+        optional_fields (List[str]): List of schema fields which are defined as
```
roll: If we look for a list of strings than we need to use something like `optional_field_names` to prevent confusion with the `field` concept (an object/dict)

PR #318, opened 2019-11-25
Fix header checks

### LINKED ISSUE #312: None of header validators(inbuilt and custom that has context="head") are invoked if there is a blank header

roll:
# Overview

- fixes #312
- fixed #314's tests


roll:
@SrividhyaGuru
Please take a look. I've updated the behavior to fix the tests. Does it solve the initial issue?

I see that the `missing-header/value` meaning is not obvious but, for now, we need to follow the spec -https://github.com/frictionlessdata/data-quality-spec

Hopefully, we will simplify things regarding these errors for `goodtables@3`

SrividhyaGuru:
@roll had a look at the changes. Works fine 👍 Thanks.

geekyrajshri:
@roll 
When can we expect this to be merged?

roll:
@SrividhyaGuru 
Releasing as `goodtables@2.4.1`

SrividhyaGuru:
Thanks @roll 👍

PR #343, opened 2020-03-18
Fixed exception thrown when a field has a wrong type in a unique column

### LINKED ISSUE #342: Cell with wrong type in unique column causes an exception (type_or_format_error & unique_field checks combination)
# Overview

There is an error caused by the combination of the checks `type_or_format_error`and `unique_field` in certain cases.

When validating the same data, but changing the rows order, returns the report without throwing an exception.

To reproduce the bug with the same script, it's important to note the `"type": "integer"` in the column marked as `"unique": True`

Here is a simple, self contained, script to demonstrate the problem:
```python
# !/usr/bin/python
# -*- coding: utf-8 -*-

'''
Bug throws an exception when a cell has a wrong type/format in a unique column (goodtables@2.4.8)
-------------------------------------------------------------------------------------------------
throws:
File ".../python3.7/site-packages/goodtables/checks/unique_constraint.py", line 39, in check_row
    row_number = column_cells[0]['row-number']
IndexError: tuple index out of range
'''

import json
from goodtables import validate


data_noexception = [
    ["id", "unique_number"],
    ["a2", "no-number"],
    ["a1", 100],
]
# same as 'data_valid' but different row order
data_exception = [
    ["id", "unique_number"],
    ["a1", 100],
    ["a2", "no-number"]
]

# descriptor template with schema to insert data
DESCRIPTOR = {
    "resources": [
        {
            "name": "example",
            "data": [],
            "schema": {
                "fields": [{
                        "name": "id",
                        "type": "string"
                    },
                    {
                        "name": "unique_number",
                        "type": "integer",
                        "constraints": {
                            "unique": True
                        }
                    }
                ]
            }
        }
    ]
}

# This reports a 'type-or-format-error'.
DESCRIPTOR["resources"][0]["data"] = data_noexception
print(json.dumps(validate(DESCRIPTOR)["tables"][0]["errors"], indent=4))

# This throws an unexpected exception, but should report a 'type-or-format-error' as well.
DESCRIPTOR["resources"][0]["data"] = data_exception
print(json.dumps(validate(DESCRIPTOR)["tables"][0]["errors"], indent=4))
```

---

didiez:
# Overview
This fixes #342 but feels like a hack. I'm sure there is a better and elegant way to solve this.

---

### DIFF
--- a/goodtables/checks/unique_constraint.py
+++ b/goodtables/checks/unique_constraint.py
@@ -35,36 +35,37 @@ class UniqueConstraint(object):
                 for _, cell in enumerate(cells, start=1)
                 if cell["column-number"] in column_numbers
             )
-            column_values = tuple(cell.get('value') for cell in column_cells)
-            row_number = column_cells[0]['row-number']
+            if column_cells:
+                column_values = tuple(cell.get('value') for cell in column_cells)
+                row_number = column_cells[0]['row-number']
 
-            all_values_are_none = (set(column_values) == {None})
-            if not all_values_are_none:
-                if len(cache[column_values]) > 0:
-                    if len(cache[column_values]) <= 5:
-                        message_substitutions = {
-                            "row_numbers": ", ".join(
-                                map(str, cache[column_values] + [row_number])
-                            ),
-                        }
-                    else:
-                        message_substitutions = {
-                            "row_numbers": "{rows} and {count} others".format(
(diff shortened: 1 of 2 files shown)


roll:
Thanks @didiez,

could you please add a test to showcase the fixed problem?

didiez:
> Thanks @didiez,
> 
> could you please add a test to showcase the fixed problem?

Done :)

roll:
Thanks a lot!

PR #344, opened 2020-03-19
Improved foreign key error

didiez:
# Overview

Improved 'foreign-key' error, if not caused by a composite foreign key, adding the cell causing the violation to improve the error details with the column-number

---


didiez:
Sorry for the commit history mess. I forgot about my last pull request :S

PR #346, opened 2020-03-19
Improved 'foreign-key' error with column-number, if not composite key

didiez:
# Overview

Improved 'foreign-key' error, if not caused by a composite foreign key, adding the cell causing the violation to improve the error details with the column-number

---


didiez:
Thank you @roll for this awesome project.
Keep up the good work!

roll:
Thanks @didiez and thanks for your contribution!

PR #367, opened 2020-04-27
Fix the `duplicate-row` collision bug

### LINKED ISSUE #329: Duplicate row bug
# Overview

When running the validate on the Python3.7 (32 bit) environment it produces duplicate row errors which are incorrect. On the 64 bit python the validation runs fine, as it should.

The file has 160 000 rows.
You can use my schema to create a file : https://www.medicalschemes.com/xmlschemas/dddr/2019/JsonDDDRPartYearReturnTypeCC3_2019_8.json 
 validate(r'**file', schema=r'**schema',error_limit=30,row_limit=1000000)

---

roll:
- fixes #329

### DIFF
--- a/goodtables/checks/duplicate_row.py
+++ b/goodtables/checks/duplicate_row.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import six
 import json
 from ..registry import check
 from ..error import Error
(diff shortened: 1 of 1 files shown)


goodtables/checks/duplicate_row.py
```diff
@@ -24,7 +24,9 @@ def check_row(self, cells):
 
         # Get pointer
         try:
-            pointer = hash(json.dumps(list(cell.get('value') for cell in cells)))
+            values = list(cell.get('value') for cell in cells)
+            # https://github.com/frictionlessdata/goodtables-py/issues/329
+            pointer = hash(json.dumps(values + [len(values)]))
```
davidpeckham: @roll, I don't think this will avoid collisions.

```values + [len(values)]```

Since rows often have the same number of values.

roll: @davidpeckham 
Sure, thanks. I've fixed it already :smi

roll:
@davidpeckham 
It seems I make the quick fix work but I'll def re-iterate on it later probably switching to other hash mechanism

PR #377, opened 2020-07-01
Fixed an exception that occurred when validating a table with a column that has a non matching header

### LINKED ISSUE #376: Column with non matching header causes an exception when order_fields is set to True
# Overview

Validating a table with a column whose header name is not included in the schema causes an exception/crash.

## System and versions
* goodtables version 2.5.0
* Python 3.8.2
* macOS 10.15.5

## Example

Here is a example that demonstrates the issue:

```python
from goodtables import validate

def main():
    data_exception = [
        ["Name", "Type", "Extra"],
        ["A", "Type A", "extra info A"],
        ["B", "Type B", "extra info B"],
    ]

    descriptor = {
        "resources": [
            {
                "name": "example",
                "data": data_exception,
                "schema": {
                    "fields": [{
                        "name": "Name",
                        "type": "string"
                    },
                        {
                            "name": "Type",
                            "type": "string"
                        }
                    ]
                }
            }
        ]
    }

    validate(descriptor, order_fields=True)


if __name__ == '__main__':
    main()

```

## Possible solution

The issue seems to be originated in the `non_matching_header.py` file. In line 62 (version 2.5.0 of goodtables):

```python
new_cell = create_cell(header=None, field=cell.get('field'))
```

The variable `cell` is being accessed without having been previously declared.

I believe that a valid fix would simply be to declare it right before:
```python
cell = cells[index]
```

---

ajpallares:
Fixes #376 

---

### DIFF
--- a/goodtables/checks/non_matching_header.py
+++ b/goodtables/checks/non_matching_header.py
@@ -59,6 +59,7 @@ def _check_with_ordering(cells):
             # which means there is no matching field for the header.
             # Add a new cell with an empty header so that the missing-header
             # error will be thrown later.
+            cell = cells[index]
             new_cell = create_cell(header=None, field=cell.get('field'))
             new_headers.append(new_cell)
             # Change the current cell's field to None


roll (changes requested):
Hi, thanks!

Could you please take a look at the failing tests - https://travis-ci.org/github/frictionlessdata/goodtables-py/jobs/703890125. Do they also have to be fixed or we need to update the solution?

PR #509, opened 2020-11-06
add @pytest.mark.ci to two SQL tests

chris48s:
# Overview

These two tests are missing the `@pytest.mark.ci` decorator.
Because the test suite runs in a pre-commit hook this means I either have to have a local test DB set up or make every commit with `--no-verify` (somewhat negating the point of the hook)

---


roll (approved):
Thanks!

PR #512, opened 2020-11-09
Stop sorting keyed sources as the order is now guaranteed by Python

### LINKED ISSUE #513: Stop sorting keyed sources as the order is now guaranteed by Python
# Overview

> Standard dict objects preserve order in the reference (CPython) implementations of Python 3.5 and 3.6, and this order-preserving property is becoming a language feature in Python 3.7.

http://gandenberger.org/2018/03/10/ordered-dicts-vs-ordereddict/

roll:
- fixes #513

> Standard dict objects preserve order in the reference (CPython) implementations of Python 3.5 and 3.6, and this order-preserving property is becoming a language feature in Python 3.7.

http://gandenberger.org/2018/03/10/ordered-dicts-vs-ordereddict/


roll:
Hi @chris48s,

Please take a look. We have discussed this one

Issue #475, opened 2020-10-13
Integration with CKAN as a data portal


roll:
# Overview

An important step for Frictionless Framework is to provide an ability to read and write packages from different data portals (CKAN/Github/Zenodod/etc) so the users can publish and access their packages easily and using a straightforward API. This issue is for CKAN integration. The implementation is already prototyped in v5 branch.

### Read package

Read package from a dataset that has a datapackage.json/yaml:

```python
package = Package("https://dados.gov.br/dataset/consumo-de-energia-orgaos-publicos")
package = Package.from_ckan(...) # alias
```

Read package from a repository without a datapackage.json/yaml. We probably need to filter files and add only CSV/XLS(X) to the package. Also CkanControl should have this configurable. We need to map as much as possible metadata provided by CKAN (for now, using `frictionless-ckan-mapper`):

```python
package = Package("https://demo.ckan.org/dataset/sample-dataset-1")
package = Package("<link>", control=portals.CkanControl(formats=['csv']))
```

### Write package

Publish a package on CKAN (for now, only if the dataset doesn't exist). Also we need to provide an ability to store credentials in ENV/etc. We need to map as much as possible metadata provided by Package:

```python
package.to_ckan()
```

### Read catalog

Read catalog from ckan search. Design some search configurations like `limit` and `offset` (pagination).

```python
catalog = Catalog(control=portals.CkanControl(baseurl=<ckan-instance-url>, search="<frictionless>")
for package in catalog.packages:
  print(package.name)
```

## Plan

- [ ] prototype the functionality based on the functional requirements
- [ ] get feedback from @roll on the implementation
- [ ] finish the implementation
- [ ] design the testing approach (probably using pytest.vcr fro reading/from what instance? but how to test writing?) (sync with Adria on the best practices of testing against CKAN instances)
- [ ] write a great deal of tests to be sure that the integration works correctly
- [ ] write a comprehensive tutorial - https://framework.frictionlessdata.io/docs/tutorials/tutorials-overview (new section Portals Tutorials)

rufuspollock:
@roll with the new work on https://github.com/frictionlessdata/frictionless-ckan-mapper this should be easier.

Issue #413, opened 2020-04-03
Improve naming mapping for ckan integration


roll:
# Overview

For now, it uses underscored resource ids. It would be much better if it extracts resource names.

roll:
MERGED into #393

Issue #447, opened 2020-10-01
Clean up the codebase


roll:
# Tasks

- [x] process "NOTE"s
- [x] move big data files to some other repo
- [x] recover skipped tests
- [ ] Consider `loader.byte_sample/text_sample` and proper sample caching/chain (less requests)
- [ ] Unify source detection like Resource(source)
- [ ] Deprecate Table/File
- [ ] Merge Table/File tests to Resource
- [ ]  Add from/to_csv/etc aliases to resource?
- [ ] Remove `validate_inqury` (and use `validate_package` instead)?
- [ ]  Reshape control/dialect/query/infer (header to query/detect_encoding?)
- [ ]  Review outstanding NOTE/TODOs in the repo
- [ ]  Sync checks/steps and/with error/types!?
- [ ]  Finalize error group names; Сheck's API
- [ ]  Make documentation hierarchical (use .build/tutorials)?
- [ ]  Plugins priority (or prepare/create_resource)!
- [ ]  Rework extract/validate API accepting resource/package??
- [ ]  Report should have resource property for tables instead of schema/dialect/etc?
- [ ] process "TODO"s
- [ ] Remove deprecated code for version 4 (table/file)
- [ ]  Add Package/Resource/Schema/Field guides
- [ ]  Remove notebook sources from git?
- [ ]  Fix coverage to codecov sending?
- [ ] coverage -> 90%

roll:
DONE in #607 #613 #621 and others

PR #528, opened 2020-11-13
Migrate CKAN plugin

### LINKED ISSUE #393: Migrate CKAN plugin
# Overview

For the CKAN plugin, we need to write a `plugins.ckan.CkanStorage class` which already exists in the repository but we need to update to the latest `Storage` interface requirements:

```
class Storage:
    __init__(self, **options):
    __iter__(self):
   read_resource(self, name, **options)
   read_package(self, **options)
   write_resource(self, resource, *, force=False, **options)
   write_package(self, package, *, force=False, **options)
   delete_resource(self, name, *, ignore=False, **options)
   delete_package(self, names, *, ignore=False, **options)
```

We already have implemented:
- `plugins.sql.SqlStorage` (more complex as it supports foreign keys)
- `plugins.pandas.PandasStorage`

So it should be relatively easy to update `CkanStorage` based on the reference interface and 2 existent implementations.

We also need:
- update `ckanPlugin.create_storage`
- add `resource.from_ckan/to_ckan`
- add `package.from_ckan/to_ckan`

For testing we use a reference package that we should be able to export and then import back validating results, for example, for pandas:

```python
def test_storage_types():

    # Export/Import
    source = Package("data/storage/types.json")
    storage = source.to_pandas()
    target = Package.from_pandas(dataframes=storage.dataframes)

    # Assert metadata
    assert target.get_resource("main").schema == {
        "fields": [
            {"name": "any", "type": "string"},  # type fallback
            {"name": "array", "type": "array"},
            {"name": "boolean", "type": "boolean"},
            {"name": "date", "type": "date"},
            {"name": "date_year", "type": "date"},  # format removal
            {"name": "datetime", "type": "datetime"},
            {"name": "duration", "type": "duration"},
            {"name": "geojson", "type": "object"},
            {"name": "geopoint", "type": "array"},
            {"name": "integer", "type": "integer"},
            {"name": "number", "type": "number"},
            {"name": "object", "type": "object"},
            {"name": "string", "type": "string"},
            {"name": "time", "type": "time"},
            {"name": "year", "type": "integer"},  # type downgrade
            {"name": "yearmonth", "type": "array"},  # type downgrade
        ],
    }

    # Assert data
    assert target.get_resource("main").read_rows() == [
        {
            "any": "note1",
            "array": ["Mike", "John"],
            "boolean": True,
            "date": datetime.date(2015, 1, 1),
            "date_year": datetime.date(2015, 1, 1),
            "datetime": datetime.datetime(2015, 1, 1, 3, 0),
            "duration": isodate.parse_duration("P1Y1M"),
            "geojson": {"type": "Point", "coordinates": [33, 33.33]},
            "geopoint": [30, 70],
            "integer": 1,
            "number": 7,
            "object": {"chars": 560},
            "string": "good",
            "time": datetime.time(3, 0),
            "year": 2015,
            "yearmonth": [2015, 1],
        },
    ]

    # Cleanup storage
    storage.delete_package(target.resource_names)


def test_storage_integrity():

    # Export/Import
    source = Package("data/storage/integrity.json")
    storage = source.to_pandas()
    target = Package.from_pandas(dataframes=storage.dataframes)

    # Assert metadata (main)
    assert target.get_resource("main").schema == {
        "fields": [
            # added required
            {"name": "id", "type": "integer", "constraints": {"required": True}},
            {"name": "parent", "type": "number"},  # type downgrade
            {"name": "description", "type": "string"},
        ],
        "primaryKey": ["id"],
        # foreign keys removal
    }

    # Assert metadata (link)
    assert target.get_resource("link").schema == {
        "fields": [
            # added required
            {"name": "main_id", "type": "integer", "constraints": {"required": True}},
            # added required; removed unique
            {"name": "some_id", "type": "integer", "constraints": {"required": True}},
            # removed unique
            {"name": "description", "type": "string"},
        ],
        "primaryKey": ["main_id", "some_id"],
        # foreign keys removal
    }

    # Assert data (main)
    assert target.get_resource("main").read_rows() == [
        {"id": 1, "parent": None, "description": "english"},
        {"id": 2, "parent": 1, "description": "中国人"},
    ]

    # Assert data (link)
    assert target.get_resource("link").read_rows() == [
        {"main_id": 1, "some_id": 1, "description": "note1"},
        {"main_id": 2, "some_id": 2, "description": "note2"},
    ]

    # Cleanup storage
    storage.delete_package(target.resource_names)


def test_storage_constraints():

    # Export/Import
    source = Package("data/storage/constraints.json")
    storage = source.to_pandas()
    target = Package.from_pandas(dataframes=storage.dataframes)

    # Assert metadata
    assert target.get_resource("main").schema == {
        "fields": [
            {"name": "required", "type": "string"},  # constraint removal
            {"name": "minLength", "type": "string"},  # constraint removal
            {"name": "maxLength", "type": "string"},  # constraint removal
            {"name": "pattern", "type": "string"},  # constraint removal
            {"name": "enum", "type": "string"},  # constraint removal
            {"name": "minimum", "type": "integer"},  # constraint removal
            {"name": "maximum", "type": "integer"},  # constraint removal
        ],
    }

    # Assert data
    assert target.get_resource("main").read_rows() == [
        {
            "required": "passing",
            "minLength": "passing",
            "maxLength": "passing",
            "pattern": "passing",
            "enum": "passing",
            "minimum": 5,
            "maximum": 5,
        },
    ]

    # Cleanup storage
    storage.delete_package(target.resource_names)


# NOTE: can we add consratins support to Pandas?
@pytest.mark.skip
@pytest.mark.parametrize(
    "field_name, cell",
    [
        ("required", ""),
        ("minLength", "bad"),
        ("maxLength", "badbadbad"),
        ("pattern", "bad"),
        ("enum", "bad"),
        ("minimum", 3),
        ("maximum", 9),
    ],
)
def test_storage_constraints_not_valid_error(field_name, cell):
    pass


def test_storage_read_resource_not_existent_error():
    storage = PandasStorage()
    with pytest.raises(exceptions.FrictionlessException) as excinfo:
        storage.read_resource("bad")
    error = excinfo.value.error
    assert error.code == "storage-error"
    assert error.note.count("does not exist")


def test_storage_write_resource_existent_error():
    resource = Resource(path="data/table.csv")
    storage = resource.to_pandas()
    with pytest.raises(exceptions.FrictionlessException) as excinfo:
        storage.write_resource(resource)
    error = excinfo.value.error
    assert error.code == "storage-error"
    assert error.note.count("already exists")
    # Cleanup storage
    storage.delete_package(list(storage))


def test_storage_delete_resource_not_existent_error():
    storage = PandasStorage()
    with pytest.raises(exceptions.FrictionlessException) as excinfo:
        storage.delete_resource("bad")
    error = excinfo.value.error
    assert error.code == "storage-error"
    assert error.note.count("does not exist")
```

Every plugin has its own limitations so in the code snippet above we note type fallbacks/downgrades and metadata removals.

# References

- how it works in the prev software generation - https://frictionlessdata.io/how-to/how-to-use-data-packages-with-ckan/ (note TODO in step 2)
- https://github.com/frictionlessdata/frictionless-py/issues/413 (an issue created after the mentioned TODO)
- https://github.com/frictionlessdata/datapackage-pipelines-ckan-driver
- see "Contributing" and other tech documents in the root of the repo

# Tasks

- [ ] migrating the plugin
- [ ] fixing the naming problem from the original implementation
- [ ] write a guide

chris48s:
# Overview

Closes #393
Refs #475

* For this PR I have only looked at achieving feature parity with https://github.com/frictionlessdata/datapackage-pipelines-ckan-driver (accepting that the new API interface doesn't allow for _exact_ feature parity). I haven't attempted to address https://github.com/frictionlessdata/frictionless-py/issues/413 or other package/resource metadata concerns. I guess that is where https://github.com/frictionlessdata/frictionless-ckan-mapper comes into play? Lets deal with that under #475
* With the tests, we couldn't do quite the same 'round-trip' trick of reading test data into a package, writing it to a storage and then reading back into a package again because CKAN is an external service and we need to mock our interactions with it. For the read tests, I've based them pretty heavily on the test data from https://github.com/frictionlessdata/datapackage-pipelines-ckan-driver instead of creating new test data. For the write tests, I've used the frictionless storage test data and made assertions on the JSON payloads we're POSTing to the mock CKAN endpoints.

I'll leave some more notes inline on the diff.

---


frictionless/resource.py
```diff
@@ -635,6 +635,24 @@ def from_storage(storage, *, name):
         """
         return storage.read_resource(name)
 
+    @staticmethod
+    def from_ckan(*, base_url, resource_id, api_key=None):
```
chris48s, line 639: * `dataset_id` is required for `Package.from_ckan`
* For `Package.to_ckan` and `Resource.to_ckan`, `dataset_id` is not required, but providing it allows us to reduce the search space (so we need to make fewer calls to `datastore_search` if we have it)
* For `Resource.from_ckan`, `dataset_id` would do nothing, so I haven't bothered allowing `dataset_id` as a param to this method

frictionless/plugins/ckan.py
```diff
@@ -31,7 +34,8 @@ def create_parser(self, resource):
         pass
 
     def create_storage(self, name, **options):
-        pass
+        if name == "ckan_datastore":
```
chris48s, line 37: I was in two minds about whether to make this `ckan` or `ckan_datastore`.

frictionless/plugins/ckan.py
```diff
@@ -230,32 +223,61 @@ def write_table_convert_field_type(self, type):
             "object": "json",
             "array": "text[]",
             "geojson": "json",
-            "date": "date",
+            "date": "text",
```
chris48s, line 226: I made this `"date": "text",` instead of `"date": "date",` because I was hitting
`The data was invalid: invalid input syntax for type date: "2015"`
on "date_year" in the types fixture if I used `date` (although "date" from the fixture does work if I set `"date": "date",` here).
I'm not sure what the right thing to do here is?
- What I've done here (`"date": "text",`)
- Special-case `date` somewhere if the `format` param is anything other than `Y-m-d` format
- Is this actually a bug in the date type/write_cell ?

I don't think I understand well enough what you're trying to achieve clearly enough to make a judgement on it.

frictionless/plugins/ckan.py
```diff
@@ -230,32 +223,61 @@ def write_table_convert_field_type(self, type):
...
-    def write_table_row_stream(self, name, row_stream, method="upsert"):
-        table = self.read_table(name)
+    def _write_table_row_stream(self, name, row_stream):
+        table = self._read_table(name)
         datastore_upsert_url = "{}/datastore_upsert".format(self.__base_endpoint)
-        records = []
-        for r in row_stream:
-            #  r = self.__mapper.convert_row(r, table.eschema)
-            records.append(r)
+        records = [r.to_dict(json=True) for r in row_stream]
```
chris48s, line 236: TODO: We're basically jamming the whole dataset into a single JSON payload here. From a practical perspective any real-life CKAN instance will have a limit on the maximum size of an HTTP POST body that it will accept so for large datasets this operation is going to fail.
This is copied from the https://github.com/frictionlessdata/datapackage-pipelines-ckan-driver implementation so it doesn't represent a regression, but as a future piece of work we should really give the option to split the dataset into chunks here for dealing with large datasets. Suggest logging under #475

frictionless/plugins/ckan.py
```diff
@@ -230,32 +223,61 @@ def write_table_convert_field_type(self, type):
...
+        if name in self.__tables:
+            del self.__tables[name]
+
+        # Invalidate cache
+        self.__bucket_cache = None
+
+    def delete_resource(self, name, *, ignore=False):
+        self._write_table_remove(name, ignore=ignore)
+
+    def delete_package(self, names, *, ignore=False):
```
chris48s, line 277: Bit of a tangent, but I find it really surprising that `write_package()` takes a `Package` object but `delete_package()` takes a list of resource names.
Any reason for that design decision? Would it make sense to change the storage API so `delete_package()` takes a `Package` object ahead of the 4.0 release?

tests/plugins/test_ckan.py
```diff
@@ -0,0 +1,585 @@
...
+                {"id": "duration", "type": "text"},  # type fallback
+                {"id": "geojson", "type": "json"},
+                {"id": "geopoint", "type": "text"},  # type fallback
+                {"id": "integer", "type": "int"},
+                {"id": "number", "type": "float"},
+                {"id": "object", "type": "json"},
+                {"id": "string", "type": "text"},
+                {"id": "time", "type": "time"},
+                {"id": "year", "type": "int"},
+                {"id": "yearmonth", "type": "text"},  # type fallback
```
chris48s, line 354: We're falling back `any`, `duration`, `geopoint` and `yearmonth` to `text` here. Do all of those seem right?

chris48s:
I've pinned responses to 0.12.0 in 2b08bf5 to get the tests to pass. They were failing because of https://github.com/getsentry/responses/issues/358 / https://github.com/spulec/moto/issues/3460
Hopefully there will be a fix for that soon and we can update it.

chris48s:
Argh :exploding_head:  . Tried restarting this a few times and it the build passes on python 3.6 and 3.7 but the python 3.8 build (and only python 3.8) consistently hangs/times out on `tests/validate/test_inquiry.py`, which is not modified in this PR. Just looking at it, there's been a couple of commits to master since the last passing build so I'm not 100% sure if that is new in this PR. I'll see if I can work it out next week.

roll:
No worries Chris i haven't managed to make tests fully stable

roll:
Thanks, Chris, it's a great work :+1: 

I'm going to merge it as it is and I'll go through it more closely (based on your comment in this PR) when I'm on the last bits for the release in #447

PR #540, opened 2020-11-22
Improve "Working with" guides

roll:
# Overview

Adding more text and examples


lwinfree:
These are so nice @roll

PR #622, opened 2021-01-22
read_tow_stream is read_rows in recent version

pedrovgp:
In the version 3.48 the method read_row_stream() does not exist. read_rows works though.

---


roll (approved):
Thank you!

PR #623, opened 2021-01-22
Expliciting the relationship

pedrovgp:
This is a suggestion. The first time I read, it was not clear that neighbor_id was a self reference to id.

---


roll (approved):
Thanks! Good catch

Issue #616, opened 2021-01-12
Validation fails with a compression-error


ethho:
# Overview

Commits 1528d00f822551e28be969195ec75499d6c12496 and more recent, validation using CLI fails: 
```bash
$ frictionless validate --nopool datapackage.json
---
invalid: datapackage.json
---
=================  =========================================================================================
code               message
=================  =========================================================================================
compression-error  The data source could not be successfully decompressed: compression "no" is not supported
=================  =========================================================================================
```

Validation passes without errors on 5bd794179dcbaab080e1e0d0e692b6c875e44d43.

---

Please preserve this line to notify @roll (lead of this repository)

roll:
@eho-tacc,

Thanks for the quick feedback I will release a fix soon

roll:
FIXED in `frictionless@4.0.0a7`

lwinfree:
yes it is working for me now!

PR #625, opened 2021-01-23
adds compression to describe

lwinfree:
quick fix for a bug in the docs. @roll I assume there is a better fix here?

### DIFF
--- a/docs/guides/introduction-guide.md
+++ b/docs/guides/introduction-guide.md
@@ -85,6 +85,7 @@ Let's update our metadata and save it to the disc:
 from frictionless import describe
 
 resource = describe("data/countries.csv", infer_missing_values=["", "n/a"])
+resource.compression = 'no'
 resource.schema.get_field("neighbor_id").type = "integer"
 resource.schema.foreign_keys.append(
     {"fields": ["neighbor_id"], "reference": {"resource": "", "fields": ["id"]}}


roll:
Hi @lwinfree,

I don't think this line is needed. I checked both v3/v4 and it works for me without it. Can you please share the error that happened?

roll:
I think it was fixed here - https://github.com/frictionlessdata/frictionless-py/issues/616

Please re-open if it's still a problem

lwinfree:
yes it is working for me now! thx!

lwinfree:
wait, actually it still isn't working (sorry I tried something else that did work & got confused).

In this section (https://framework.frictionlessdata.io/docs/guides/introduction-guide#describing-data) of the intro guide, after: 
```
Let's update our metadata and save it to the disc:

from frictionless import describe

resource = describe("data/countries.csv", infer_missing_values=["", "n/a"])
resource.schema.get_field("neighbor_id").type = "integer"
resource.schema.foreign_keys.append(
    {"fields": ["neighbor_id"], "reference": {"resource": "", "fields": ["id"]}}
)
resource.to_yaml("tmp/countries.resource.yaml")
```
the `data/countries.csv` has `compression: ''` instead of `compression: 'no'`

This then causes an error when we try to extract the file later in the extract section:
```
frictionless extract tmp/countries.resource.yaml --basepath .
```
```
[compression-error] The data source could not be successfully decompressed: compression "" is not supported
```
If you manually add `resource.compression = 'no'` then you don't get a compression error.

roll:
Lilly, I've tried on a fresh `frictionless@3.48` installation:

```
$ frictionless describe data/countries.csv
---
metadata: data/countries.csv
---

compression: 'no'
compressionPath: ''
control:
  newline: ''
dialect:
  headerRows:
    - 2
encoding: utf-8
format: csv
hashing: md5
name: countries
path: data/countries.csv
query: {}
schema:
  fields:
    - name: id
      type: integer
    - name: neighbor_id
      type: string
    - name: name
      type: string
    - name: population
      type: string
scheme: file
stats:
  bytes: 136
  fields: 4
  hash: b0481536cb4ab3e5db64f0feede627fa
  rows: 5
```

My guess that an error can only happen if you run describe by frictionless@4 and then use frictionless@3. But within a version compression should be consistent

I dropped IPython to give us the ability to work with plain markdown, also it was really slow but I think we still need a way to run/test documentation. I have a few ideas so will be on it soon because it's almost not possible to test all our docs manually

lwinfree:
closing as I was sneakily actually running version 4. This works fine on version 3.8!

PR #629, opened 2021-01-25
Adds --version call / comment

hbruch:
Replaces hyphen in install fragment by hash, so it could be executed as is.
Adds --version call to a) implicitly document which version this documentation refers to, b) relate to the semver notice above
---


roll (approved):
:+1:

PR #630, opened 2021-01-25
Adds link to repository

hbruch:
For people accessing this tutorial via frictionlessdata.github.io, it might not immediately be clear, which repository is meant by 'this'.

I assume, that people working with frictionless have sufficient knowledge to perform a checkout of this repository to access the data.

Note: working from the checked out repository directory will operate on current master, not the version installed via pip install.
This could break some tutorial examples, as is currently the case for 4.x-master and 3.48.0 stable release.

---

### DIFF
--- a/docs/guides/getting-started.md
+++ b/docs/guides/getting-started.md
@@ -54,7 +54,7 @@ $ frictionless transform --help
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


roll (approved):
Good catch. Now it's not obvious when the docs not on Githib

PR #685, opened 2021-02-11
Intro edits

lwinfree:
Hi @roll & @sapetti9! Here are my proposed changes to the intro guide, getting started guide, and also the introduction. Here is the general overview:

- I added information to the [introduction](https://framework.frictionlessdata.io/docs/introduction/introduction) to give a more thorough overview of the project. Sara will be creating some graphics to replace the user story bullet point list.

- I renamed "Getting Started" to "Quick Start" (I think this makes more sense). I added some text about installation trouble shooting. Maybe we can add a link to the FAQ once we make that later.

- I renamed "Introduction Guide" to "Overview Example". I think this makes more sense as it is an example showing all of the functionality. I don't totally love this name, but I think it is better than "Introduction Guide". I edited the text at the top to give more detail about the example dataset.

What do y'all think?


lwinfree:
also, @roll if you are OK with these changes, can you change the link on pipy to point to the introduction instead of the getting started guide?

roll:
Hi @lwinfree,

It looks good!

A few comments:
- the only reason I added the "Introduction" top section is to put a copy of the readme somewhere (because README contains kinda good introduction). If we want to have a separate document instead I would drop the whole "Introduction" section then
- We have "Guides Overview" then "Quick Start" then "Overview Example". Maybe we can rephrase "Overview Examples"? Something like "First Steps", "Diving Into..", etc?

So my idea is to have as informational architecture then:
- Homepage (we can add documentation overview and some info from introduction to the homepage)
- Guides (merged with Introduction)
  - Guides Overview
  - Quick Start
  - Introduction Guide? (from the removed introduction section)
  - Overview Example (First Step? I don't know something about basics?) BTW can we just merge it with Introduction and name "Introduction Guide"?
  - Describing Data
  - Extracting Data
  - Validating Data
  - Transforming Data
  - Mastering the Framework
  - Writing an Extension

roll:
Ok, what if:

- Guides
  - Guides Overview
  - Quick Start
  - Basic Examples
  - Introduction Guide
  - Describing Data
  - Extracting Data
  - Validating Data
  - Transforming Data
  - Mastering the Framework
  - Writing an Extension

PR #707, opened 2021-02-19
adds an example, an explanation, a link, fixes a typo, etc.

augusto-herrmann:
# Overview

This is part of a review of documentation for the user feedback session. Overall I think it is excelent and I was able to follow it with ease.

A few comments on the content of the document and/or proposed changes follow.

## Style related questions

I've noticed that the documentation is not enforcing a 78 columns line width, so I have not changed any of the long lines.

Shouldn't references to classes be enclosed in code, e.g. `Resource`? I have not changed the current style of not marking it as code.

## Clarity

I was a little confused about the explanation on why the documentation says the code in [the Extracting Resource section](https://framework.frictionlessdata.io/docs/guides/extracting-data/#extracting-resource) does not make sense, but I have not changed it. It's the excerpt below:

> In many cases, the code above doesn't really make sense as we can just provide a path to the high-level extract function instead of a descriptor to the extract_resource function but the power of the descriptor is that it can contain different metadata and be stored on the disc. 

In [the Package Class section](https://framework.frictionlessdata.io/docs/guides/extracting-data/#package-class), the Python code doesn't really require us to create the data package json beforehand, so I've changed the wording in there.

## Code instructions

For the code instructions, it's not clear if the user should reuse the same Python terminal as the previous code examples, or create a new one. Some classes are repeatedly imported again, but not `pprint`, which is imported only in the first code example. I think we should choose one or the other (either a reused terminal or a new terminal) and be consistent with that choice.

---

### DIFF
--- a/docs/guides/extracting-data.md
+++ b/docs/guides/extracting-data.md
@@ -111,7 +111,7 @@ from frictionless import Resource
 
 resource = Resource('data/capital-3.csv')
 resource.infer()
-resource.schema.missing_values.append('3')
+resource.schema.missing_values.append('3') # will interpret 3 as a missing value
 resource.to_yaml('tmp/capital.resource.yaml')
 ```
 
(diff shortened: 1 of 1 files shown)


roll (approved):
Thanks!

PR #708, opened 2021-02-19
Suggestions for improvements to documentation Jen

jen-thomas:


docs/guides/extracting-data.md
```diff
@@ -2,7 +2,7 @@
 title: Extracting Data
 ---
 
-Extracting data means reading tabular data from some source. We can use various customizations for this process such as providing a file format, table schema, limiting fields or rows amounts, and much more. Let's see this with some real files:
+Extracting data means reading tabular data from a source. We can use various customizations for this process such as providing a file format, table schema, limiting field or row values, and much more. Let's see this with some real files:
```
augusto-herrmann, line 5: I might be wrong, but I think the text means limiting which fields or how many rows are to be extracted, not really their values.

docs/guides/extracting-data.md
```diff
@@ -73,22 +73,22 @@ pprint(rows)
 
 The high-level interface for extracting data provided by Frictionless is a set of `extract` functions:
 - `extract`: detects the source type and extracts data accordingly
-- `extract_package`: accepts a package descriptor and returns a map of the package's tables
 - `extract_resource`: accepts a resource descriptor and returns a data table
+- `extract_package`: accepts a package descriptor and returns a map of the package's tables
```
augusto-herrmann, line 77: Good. This order is consistent with the order of sections on this page. :+1:

docs/guides/extracting-data.md
```diff
@@ -104,16 +104,26 @@ pprint(rows)
  Row([('id', 5), ('name', 'Rome')])]
 ```
 
-In many cases, the code above doesn't really make sense as we can just provide a path to the high-level `extract` function instead of a descriptor to the `extract_resource` function but the power of the descriptor is that it can contain different metadata and be stored on the disc. Let's extend our example:
+Using the `extract_resource` function though, we can extract the resource from a descriptor. The power of the descriptor is that it can contain different metadata and be stored on the disc.
```
augusto-herrmann, line 107: Agreed. I too was a bit confused by the wording in the original there.

docs/guides/extracting-data.md
```diff
@@ -104,16 +104,26 @@ pprint(rows)
...
 resource.to_yaml('tmp/capital.resource.yaml')
 ```
+This description can then be used to extract the resource:
+
+```python
+from frictionless import extract_resource
+
+data = extract_resource('tmp/capital.resource.yaml')
+```
+This can also be done on the command-line:
```
augusto-herrmann, line 126: Adding this example is pretty useful indeed!

docs/guides/extracting-data.md
```diff
@@ -190,17 +200,27 @@ for path, rows in data.items():
...
 ```
+We can also extract the package from a descriptor using the `extract_package` function: 
+
+```python
+from frictionless import extract_package
+
+package  = extract_package('tmp/country.package.yaml')
+
+pprint(package)
+```
```
augusto-herrmann, line 219: Excellent! :+1:

augusto-herrmann (approved):
The changes proposed here do improve this page by making the wording clearer and adding useful examples.

roll (approved):
Thank you @jen-thomas and @augusto-herrmann!

I will merge as it is and update the first edit (the correct idea was spotted by @augusto-herrmann)

PR #710, opened 2021-02-22
fix sample code errors

mccalluc:
# Overview

`self` is undefined. Also: fix indentation and remove unused import.

---


roll (approved):
Thank you!

PR #729, opened 2021-03-01
Review and correct some language + use backticks for commands

peterdesmet:
# Overview

I am reviewing the transformation guide, but it referenced the introduction and quick start, so I went through some (mostly language) edits there too.

For the 4 main commands, I use `backticks` in the text to make clear that these are commands.

---


peterdesmet:
Ping @lwinfree

lwinfree (changes requested):
Thanks @peterdesmet! This looks great. There are 2 small changes I am requesting:

- in the intro doc, line 53, please change `simple` to `short` as we are trying to remove words like simple from the docs (which I didn't even notice until I was reading your edits, so thanks!)
- in the quick start doc, line 16, `their` should be `there`

Could you please make these changes and then I will merge?

Thanks - we appreciate your help here!

peterdesmet:
@lwinfree good catch! I have made the requested updates.

lwinfree (approved):
Thanks so much for these edits!

PR #732, opened 2021-03-02
Possibility to read tables with different schemas

rbngz:
# Overview

The current code of the sql plugin assumes that all tables within a storage use identical schemas. When reading a table, the code appends the name of the table to the namespace and uses this string as a parameter for self.__metadata.tables.get().

In the storage there might be tables with different schemas. Referencing the tables with the complete name seems more plausible to me and avoids the storage-error "Resource X does not exist".

This is my first contribution ever. Please be nice 🥺
 
---


roll (changes requested):
Great thanks!

Could you please fix the tests and add a test to highlight the problem this PR fixes?

roll:
Hi @rbngz,

Have this - https://github.com/frictionlessdata/frictionless-py/issues/749#issuecomment-804923545 - helped?

rbngz:
Hello @roll,
Thank you for providing this explanation.
I have not been able to test it yet, but I would assume that this would solve our issues. :)

PR #738, opened 2021-03-02
Included constraints to sql schema

rbngz:
# Overview

Included the string length constraints from the sql database to the schema.
---


frictionless/plugins/sql.py
```diff
@@ -244,6 +244,10 @@ def __read_convert_schema(self, sql_table):
             field = Field(name=column.name, type=field_type)
             if not column.nullable:
                 field.required = True
+            if isinstance(column.type, sa.sql.sqltypes.CHAR):
```
roll, line 249: Probably it should be the same as below: ```(sa.sql.sqltypes.CHAR, sa.sql.sqltypes.VARCHAR)```?

roll (commented):
Hi, 

Thanks for your PR!

I have a concern that varchar/char length doesn't exactly mean a constraint. I would rather convert a CHECK to be a constraint if it's possible.

WDYT. I would also share it in the forum to discuss with the community

roll:
I'm closing it for now as I think that data type information should not be converted to a constraint. E.g. if there were a `maxLength` CHECK in SQL it would become a `maxLength` constraint in Table Schema but VARCHAR's length usually doesn't mean a constraint, just storage details.

PR #743, opened 2021-03-03
Fix typos in documentation

jen-thomas:
# Overview

---


lwinfree (approved):
Thanks for these edits @jen-thomas!

PR #745, opened 2021-03-04
Review transformation guide

peterdesmet:
# Overview

Review of transformation guide. Mostly made changes to improve readability. Some things that bugged me (can be discussed during interview):

- Titles on all guides are now `Transforming Package` or `Extracting Resource`. I wonder if `Transforming a Package` or `Transform Package` wouldn't be better.
- The included examples should be more real use cases. Currently it's a bit hard to understand what is happening: the first one is a difficult concept, the second is a transformation you would never do.
- Consistent use of single or double quotes in examples.

Ping @lwinfree 

---


peterdesmet:
Here's an attempt at an easier example (capitalizing the country names), but I couldn't find in the documentation how to make it work:

```python
from pprint import pprint
from frictionless import Resource, transform, steps

# Define source resource
source = Resource(path="data/transform.csv")

# Apply transform steps
target = transform(
    source,
    steps=[
        steps.table_normalize(),
        steps.cell_convert(field_name="name", value=str.capitalize(??)), <-- how do I reference the original value?
    ],
)

# Print resulting schema and data
pprint(target.schema)
pprint(target.read_rows())
```

augusto-herrmann (approved):
I have read the changes proposed here and I do believe they do improve the clarity and readability of this page. Thanks, @peterdesmet !

lwinfree (approved):
Thanks @peterdesmet! These changes are very helpful! Evgeny will be looking at updating the examples soon too.

PR #752, opened 2021-03-12
Validating edits

lwinfree:
Hi @roll ! Here are my edits for the validating guide & transforming guide. These edits are based on feedback from the 2 user sessions. Could you please review them?

There are few things changes we still need to make to these guides. The biggest change (in my opinion) is to think about how to reorganize the content in the transform guide. Let's chat about that some more next week.


roll (approved):
:tada:

PR #753, opened 2021-03-13
remove url param from sql docs

chris48s:
# Overview

Since https://github.com/frictionlessdata/frictionless-py/pull/655 `url` is no longer a named param

---


roll (approved):
Thanks!

PR #816, opened 2021-04-22
table_aggregate description added

aborruso:
# Overview

---


roll (changes requested):
Thanks!

I think we might need to change it to something like:

> Group rows under the given `group_name` then apply aggregation functions provided as `aggregation` dictionary (see example)

https://petl.readthedocs.io/en/stable/transform.html?highlight=aggregate#petl.transform.reductions.aggregate

aborruso:
You are boss, it's ok

roll:
Thank you!

roll (approved):

PR #828, opened 2021-04-27
there is a bug in BugQuery :)

aborruso:
# Overview

A small typo
---


roll (approved):
Thanks! But at least it was fun :smile:

PR #832, opened 2021-04-28
gabriel pip install frictionless[sql] problem

gabrielbdornas:
# Overview

solution for pip install frictionless[sql] problem in zsh linux terminals
---


roll (approved):
Awesome! Thanks

PR #881, opened 2021-07-14
possible descriptor type detection

### LINKED ISSUE #874: Extract/validate needs to guess descriptor type
# Overview

Currently, this one will not work as expected:

```python
report = validate(descriptor='datapackage.yaml') # will be handled as a default (resource)
```

mor120:
# Overview

closes #874 
---
It is a possible way for detecting a type by a descriptor, if the type is not provided.
As @roll suggested, this is done by creating a file for the descriptor.

### DIFF
--- a/frictionless/validate/main.py
+++ b/frictionless/validate/main.py
@@ -26,6 +26,11 @@ def validate(source=None, type=None, **options):
         type = "package" if file.multipart else file.type
         if type == "table":
             type = "resource"
+        if "descriptor" in options:
+            file = system.create_file(
+                options.get("descriptor"), basepath=options.get("basepath", "")
+            )
+            type = file.type
     module = import_module("frictionless.validate")
     validate = getattr(module, "validate_%s" % type, None)
     if validate is None:
--- a/frictionless/validate/main.py
+++ b/frictionless/validate/main.py
@@ -30,7 +30,7 @@ def validate(source=None, type=None, **options):
             file = system.create_file(
                 options.get("descriptor"), basepath=options.get("basepath", "")
             )
-            type = file.type
+            type = file.type if file.type == "package" else "resource"
     module = import_module("frictionless.validate")
     validate = getattr(module, "validate_%s" % type, None)
     if validate is None:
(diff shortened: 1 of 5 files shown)


frictionless/validate/main.py
```diff
@@ -26,6 +26,11 @@ def validate(source=None, type=None, **options):
         type = "package" if file.multipart else file.type
         if type == "table":
             type = "resource"
+        if "descriptor" in options:
+            file = system.create_file(
+                options.get("descriptor"), basepath=options.get("basepath", "")
+            )
+            type = file.type
```
roll: I would fall back to `resource` type if the descriptor is detected to be a `table` as `file.type == 'table'`. Otherwise, something like this `validate(descriptor='table.csv')` will crash the lib (we might need to add a test for this edge case)

tests/validate/test_package.py
```diff
@@ -465,3 +465,8 @@ def test_validate_package_using_detector_schema_sync_issue_847():
         resource.detector = Detector(schema_sync=True)
     report = validate(package)
     assert report.valid
+
+
+def test_validate_package_descriptor_type():
```
roll: Can you please add a few more tests like for `resource`? And some edge cases like `validate(descriptor='table.csv')`

roll (changes requested):
Hi @mor120, thanks for helping with this issue!

Here are a few suggestions I added to the code and also:
- Can you please add a similar logic to `extract` function?

frictionless/validate/main.py
mor120: I have added a fall back to resource if the type is not a package.

Furthermore, I would like to mention that passing a ```table``` type file as a ```descriptor``` like - ```validate(descriptor='table.csv')``` won't raise any exception but will retrieve an error: 
```Python
print(validate(descriptor='table.csv').flatten())
[1, None, None, 'resource-error']
```

mor120:
Hey @roll,
Following the last commit I have made, I think we should add a test to  ```tests/extract/test_resource.py``` that will make sure that a table passed as a descriptor will be handled as a resource.

So that when calling the function:

```Python
extract(descriptor="data/package/data.csv")
```
A **resource-error** exception will be raised:

```Python
frictionless.exception.FrictionlessException: [resource-error].....
```

And not **general-error** exception:
```Python
frictionless.exception.FrictionlessException: [general-error] .....
```

What do you think? Is it necessary?

roll (approved):
Thanks @mor120!

It looks good. The `resource-errors` error I think is fine

Issue #760, opened 2021-03-18
Tests fail with new sqlalchemy@1.4


roll:
# Overview

See xfailed tests in #754

PR #882, opened 2021-07-16
SQL Where Clause

### LINKED ISSUE #628: Query builder / providing SQL clauses to the dialect
# Overview

I'm working on a very [similar project](https://gitlab.com/atviriduomenys/spinta/), which is a DEVT framework, like `frictionless-py`, but with focus on open government data.

Instead of reinventing the wheel, I decided to do an experiment integrating `frictionless-py`. So far it worked pretty well with "Describe" part, but I'm not sure if `frictionless-py` has a good support for data filtering.

As I understand currently filtering is provided by transformations, for example there is a [Filter Rows](https://frictionlessdata.github.io/frictionless-py/docs/guides/transforming-data/#filter-rows). I don't fully understand, how that works, what is `<formula>` and more important how it is implemented under the hood.

I will have to deal with very large tables, so it is important to know, that filtering happens at SQL database level, not on python level. If filtering will be done on Python, then it will take forever to complete.

Also, same thing applies to joins. Are joins performed on SQL database level, or `frictionless-py` loads everything into memory and does joins on Python side?

---

n0rdlicht:
# Overview

Allowing a `where` option to be present in the SqlDialect to filter data before retrieving it from an SQL source.

Example resource descriptor using `orderBy` and `where` clauses
```yaml
path: sqlite:///example.db
dialect:
   table: example
   orderBy: date asc
   where: date > '2019-01-29' and category = 'example'
```

fixes #628

---


roll (changes requested):
Hi @n0rdlicht,

It looks good! cc @sirex 

I think you need to update your local `sqlalchemy` to make the tests pass - https://github.com/frictionlessdata/frictionless-py/issues/760

tests/plugins/test_sql.py
```diff
@@ -13,7 +13,7 @@ def test_sql_parser(database_url):
     with Resource(database_url, dialect=dialect) as resource:
         assert resource.schema == {
             "fields": [
-                {"name": "id", "type": "integer"},
+                {"name": "id", "type": "integer", "constraints": {"required": True}},
```
roll: It's not needed. Please see the review summary comment

n0rdlicht: Should be fixed now, indeed had sqlalchemy 1.3 for some reason.

roll (approved):
Thanks!

PR #912, opened 2021-09-22
Add fields, limit, sort and filter options to CkanDialect

n0rdlicht:
# Overview

To leverage the CKAN Storage API this PR includes four new keys on the `CkanDialect`:

- `fields?` _array_ - limit ckan query to certain fields
- `limit?` _int_ - limit number of returned entries
- `sort?` _str_ - sort returned entries, e.g. by date descending: `date desc`
- `filters?` _dict_ - filter data, e.g. field with value: `{ "key": "value" }`

edit: currently just modified the cassette queries, but these probably need to be rerun against a real CKAN instance to verify no breaking behaviour. Also to include tests for the new keys.

---


roll (approved):
Thanks. It looks good!

We definitely need some real CKAN instance to rebase tests there. The demo stopped working at some point.

PR #920, opened 2021-10-01
Update basic-examples.md

mirianbr:
Change python code to be os-independent; the previous code won't run on Windows (ls call)

---


roll (approved):
:+1:

PR #923, opened 2021-10-14
Accept integer and float numbers as string in xlsx files

pierredittgen:
# Overview

Trying to validate an [xlsx file](https://git.opendatafrance.net/validata/validata-ui/uploads/dc50b16c803e2e2d61a645019cd75b3c/test1.xlsx) against a [table schema](https://git.opendatafrance.net/validata/validata-ui/uploads/dc50b16c803e2e2d61a645019cd75b3c/test1.xlsx) with `frictionless-py`, 2 errors are reported:

```bash
frictionless validate --schema https://gitlab.com/arsante/atlasante/schema-dae/-/raw/1.0.2/schema.json https://git.opendatafrance.net/validata/validata-ui/uploads/dc50b16c803e2e2d61a645019cd75b3c/test1.xlsx
# -------
# invalid: https://git.opendatafrance.net/validata/validata-ui/uploads/dc50b16c803e2e2d61a645019cd75b3c/test1.xlsx
# -------

===  =====  ==========  ======================================================================================================
row  field  code        message                                                                                               
===  =====  ==========  ======================================================================================================
  2      6  type-error  Type error in the cell "97442" in row "2" and field "com_cp" at position "6": type is "string/default"
  2     13  type-error  Type error in the cell "0" in row "2" and field "acc_etg" at position "13": type is "string/default"  
===  =====  ==========  ======================================================================================================
```
Problem is due to number values found in fields declared as string. However, the cells concerned are formatted as string.

Validating a [CSV file](https://git.opendatafrance.net/validata/validata-ui/uploads/36de6571bd2b60e53051d1256d9027f2/test1.csv) with the same content reports no errors:

```bash
frictionless validate --encoding utf-8 --schema https://gitlab.com/arsante/atlasante/schema-dae/-/raw/1.0.2/schema.json https://git.opendatafrance.net/validata/validata-ui/uploads/36de6571bd2b60e53051d1256d9027f2/test1.csv
# -----
# valid: https://git.opendatafrance.net/validata/validata-ui/uploads/36de6571bd2b60e53051d1256d9027f2/test1.csv
# -----
```

This PR adds the opportunity to accept an integer or a float as a valid string value. This solves the problem with the XLSX file and provides a consistent solution between CSV and XLSX files.

---


roll:
Hi @peterdesmet,

This behavior is intended as in-general numbers are not compatible with strings (similar to SQL/etc). I think the solution for this problem lays in the ExcelParser or in using `any` data type. WDYT?

pierredittgen:
> This behavior is intended as in-general numbers are not compatible with strings (similar to SQL/etc). I think the solution for this problem lays in the ExcelParser or in using `any` data type. WDYT?

@roll I think you're right, I will put my efforts in this direction.

pierredittgen:
Hi @roll,

I realized that enabling `preserve_formatting=True` in `ExcelDialect` resolved the problem.

As all the checks are passing, I wonder if this option is still useful and if `preserve_formatting=True` couldn't become the new default? 

What do you think?

roll:
I think we can't because it will be a breaking change

pierredittgen:
> I think we can't because it will be a breaking change

Sure @roll, I will explicitly use `preserve_formatting=True`.

Thanks for your time.

roll:
Thanks!

PR #943, opened 2021-11-08
Markdown formatting error

Grandient:
# Overview

The current format makes it look like the bullet point and the paragraph below is one and the same. These however contain two different ideas. I have added a new line to fix this formatting issue.

---


roll:
Thanks!

PR #952, opened 2021-12-02
Use row values for `row-error` cells

### LINKED ISSUE #951: Foreign key errors showing headers instead of row values
# Overview
* Version: **`v4.22.1`**

The foreign key errors returned in the report are the **header** values of the resource instead of the actual **row values**. Also, the message does not provide any hints about the problem: which is the foreign key violated? which fields? 
I would expect more details about the `foreign-key-error` alongside the other fields of the error found in the report. Does it makes sense?

frictionless-foreign-key-error-bug

# How to reproduce
Here is ZIP with a minimal example to reproduce the problem: [frictionless-foreign-key-bug.zip](https://github.com/frictionlessdata/frictionless-py/files/7633635/frictionless-foreign-key-bug.zip)
Use the **awesome** `frictionless CLI` and `jq` just to get to the errors:
```
frictionless validate datapackage.json --type package --json | jq .tasks[].errors
```
Result:
```json
[
  {
    "cells": [
      "id",
      "first_name",
      "last_name",
      "dept_id"
    ],
    "rowNumber": 8,
    "rowPosition": 9,
    "code": "foreign-key-error",
    "name": "ForeignKey Error",
    "tags": [
      "#table",
      "#row"
    ],
    "note": "not found in the lookup table",
    "message": "Row at position \"9\" violates the foreign key: not found in the lookup table",
    "description": "Values in the foreign key fields should exist in the reference table"
  },
  {
    "cells": [
      "id",
      "first_name",
      "last_name",
      "dept_id"
    ],
    "rowNumber": 18,
    "rowPosition": 19,
    "code": "foreign-key-error",
    "name": "ForeignKey Error",
    "tags": [
      "#table",
      "#row"
    ],
    "note": "not found in the lookup table",
    "message": "Row at position \"19\" violates the foreign key: not found in the lookup table",
    "description": "Values in the foreign key fields should exist in the reference table"
  }
]
[]
```

Please replace this line with full information about your idea or problem. If it's a bug share as much as possible to reproduce it

---

didiez:
fixes #951

---

### DIFF
--- a/tests/validate/test_package.py
+++ b/tests/validate/test_package.py
@@ -216,7 +216,7 @@ def test_validate_package_schema_foreign_key_internal_resource_violation():
     del descriptor["resources"][1]["data"][4]
     report = validate(descriptor)
     assert report.flatten(["rowPosition", "fieldPosition", "code", "cells"]) == [
-        [5, None, "foreign-key-error", ["4", "rio", "None"]],
+        [5, None, "foreign-key-error", ["4", "rio", ""]],
     ]
 
 
@@ -228,7 +228,7 @@ def test_validate_package_schema_foreign_key_internal_resource_violation_non_exi
         [2, None, "foreign-key-error", ["1", "london", "2"]],
         [3, None, "foreign-key-error", ["2", "paris", "3"]],
         [4, None, "foreign-key-error", ["3", "rome", "4"]],
-        [5, None, "foreign-key-error", ["4", "rio", "None"]],
+        [5, None, "foreign-key-error", ["4", "rio", ""]],
     ]
 
 
--- a/tests/validate/test_package.py
+++ b/tests/validate/test_package.py
@@ -206,8 +206,8 @@ def test_validate_package_schema_foreign_key_self_referenced_resource_violation(
     descriptor = deepcopy(DESCRIPTOR_FK)
     del descriptor["resources"][0]["data"][4]
     report = validate(descriptor)
-    assert report.flatten(["rowPosition", "fieldPosition", "code"]) == [
-        [4, None, "foreign-key-error"],
+    assert report.flatten(["rowPosition", "fieldPosition", "code", "cells"]) == [
+        [4, None, "foreign-key-error", ["3", "rome", "4"]],
     ]
 
 
@@ -215,8 +215,8 @@ def test_validate_package_schema_foreign_key_internal_resource_violation():
     descriptor = deepcopy(DESCRIPTOR_FK)
     del descriptor["resources"][1]["data"][4]
     report = validate(descriptor)
-    assert report.flatten(["rowPosition", "fieldPosition", "code"]) == [
-        [5, None, "foreign-key-error"],
+    assert report.flatten(["rowPosition", "fieldPosition", "code", "cells"]) == [
+        [5, None, "foreign-key-error", ["4", "rio", "None"]],
     ]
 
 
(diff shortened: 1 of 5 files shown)


roll:
Thanks, @didiez, good catch!

Can you please add a simple test highlighting the bug? Also, are you interested in improving the error message at https://github.com/frictionlessdata/frictionless-py/blob/main/frictionless/resource.py#L978? We can do it in this PR to fully resolve #951

didiez:
Sure, tests added :+1: 
However, while I was adding the `cells` attribute checking to the existing foreign key tests, I have noticed that the values found in the report for the `None` (empty) cells were the string `"None"`. Is there any reason to cast every cell to string?
Should we replace the current `cells=list(map(str, row.values()))` by `cells=row.cells` to preserve the cell types and the `None` values?
https://github.com/frictionlessdata/frictionless-py/blob/1b5a274bba5b736d1d7c4fc4a3eb62e64f56d1c0/frictionless/errors/row.py#L45
This also applies to `CellError.from_row(..)`
https://github.com/frictionlessdata/frictionless-py/blob/1b5a274bba5b736d1d7c4fc4a3eb62e64f56d1c0/frictionless/errors/cell.py#L78

tests/validate/test_package.py
```diff
@@ -206,29 +206,29 @@ def test_validate_package_schema_foreign_key_self_referenced_resource_violation(
...
-    assert report.flatten(["rowPosition", "fieldPosition", "code"]) == [
-        [2, None, "foreign-key-error"],
-        [3, None, "foreign-key-error"],
-        [4, None, "foreign-key-error"],
-        [5, None, "foreign-key-error"],
+    assert report.flatten(["rowPosition", "fieldPosition", "code", "cells"]) == [
+        [2, None, "foreign-key-error", ["1", "london", "2"]],
+        [3, None, "foreign-key-error", ["2", "paris", "3"]],
+        [4, None, "foreign-key-error", ["3", "rome", "4"]],
+        [5, None, "foreign-key-error", ["4", "rio", "None"]],
```
didiez: This `"None"` as string seems like a bug in [RowError.from_row](https://github.com/frictionlessdata/frictionless-py/blob/1b5a274bba5b736d1d7c4fc4a3eb62e64f56d1c0/frictionless/errors/row.py#L45). Every value in `cells` is stringified.
Should we preserve the original value without the explicit cast to string?
See [#952](https://github.com/frictionlessdata/frictionless-py/pull/952#issuecomment-987065333)

roll:
@didiez 
I would replace "None" by "". In general, the idea of cells being strings is that clients are able to show the row without handling data types but I need to review this approach. Maybe using `row.to_list(json=True)`... But I need to create a breaking release for changes like this

didiez:
The `""` (empty) string sounds good to fix this without breaking things in a bugfix release. 
As of today, we didn't need cells with types in the report, but could be useful in some scenarios.

didiez:
@roll Should we also fix the "None" from `CellError.from_row(..)` in this PR or better to open a new one?

roll:
@didiez 
Using this one is fine I think. Can you please also add the change check for this to the tests?

roll:
@didiez 
row.cells is needed to provide an original cells information. Sorry for being slightly out of the context. Can the fk bug be somewhere else?

didiez:
> @didiez row.cells is needed to provide an original cells information. Sorry for being slightly out of the context.

I am not sure I'm following. Should we use `row.values()` or `row.cells` when creating the errors? The last revert was just to keep the old behaviour of CellError/RowError.from_row. However, in `fragments/row.py` the value passed to the error is `cells`.

> Can the fk bug be somewhere else?

I did a quick search and I think this bug (shows headers as row values) was only present for `RowError` when using `from_row` method (also in every subclass)

roll:
@didiez 
I did a quick analysis and I think we need to use `cells=list(map(to_str, row.cells))` in the both `from_row` calls. The idea of the `errors.cells` is to provide original row cells while `row.values()` returns a cleaned version. Sorry for misleading you

roll:
So I think the problems we solve:
- https://github.com/frictionlessdata/frictionless-py/blob/main/frictionless/errors/row.py#L45
- https://github.com/frictionlessdata/frictionless-py/blob/main/frictionless/errors/cell.py#L78
- and inappropriate stringification of `None` (solved by your `to_str`)

roll (approved):
Great, thanks!

PR #958, opened 2021-12-10
Number of rows constraints

### LINKED ISSUE #820: Feature request: number of records constraints
# Overview

Hi,
below you have one of the official documentation yaml file. There some constraints fields related.

A lot of time I know that I know that a resource is valid only if it is made up of at least 365 rows (some annual dataset including leap years). Or that the rows shouldn't be more than 10.

I think it should be a great validate and check feature to have.

Thank you

```
dialect:
  delimiter: ;
encoding: utf-8
format: csv
hashing: md5
layout:
  headerRows:
    - 2
name: country-2
path: country-2.csv
schema:
  fields:
    - name: id
      title: Identifier
      type: integer
    - name: neighbor_id
      title: Identifier of the neighbor
      type: integer
    - name: name
      title: Name of the country
      type: string
    - constraints:
        minimum: 0
      description: According to the year 2020's data
      name: population
      title: Population
      type: integer
  foreignKeys:
    - fields:
        - neighbor_id
      reference:
        fields:
          - id
        resource: ''
scheme: file
```

---

aivuk:
- fixes #820 

---

Add new check validator based on minimum and maximum number of lines in a resource

---


roll:
Thanks @aivuk!

I'll review on Monday. 

A few comments:
- don't forget using `make test / make link / make format / etc` locally
- please rename to `table_dimenstion` check and add max fields and min fields params; and number_rows (fixed)

--- 
Here is naming reference: https://github.com/frictionlessdata/frictionless-py/issues/820#issuecomment-825521477

frictionless/checks/regulation.py
```diff
@@ -59,6 +59,80 @@ def validate_row(self, row):
...
+            yield errors.TableDimensionsError(
+                note="Current number of rows is %s, the minimum is %s"
+                % (number_rows, self.__min_rows)
+            )
+
+    # Metadata
+
+    metadata_profile = {  # type: ignore
+        "type": "object",
+        "requred": ["min_rows", "max_rows", "min_fields", "max_fields"],
```
roll: - we need to use camelCase for in-descriptor keys e.g. `minRows`
- users should be able to set only one constraint or more so we can't require all of them

frictionless/checks/regulation.py
```diff
@@ -59,6 +59,80 @@ def validate_row(self, row):
...
+    Parameters:
+       descriptor (dict): check's descriptor
+
+    """
+
+    code = "table-dimensions"
+    Errors = [errors.TableDimensionsError]
+
+    def __init__(
+        self, descriptor=None, *, min_rows=-1, max_rows=-1, min_fields=-1, max_fields=-1
```
roll: - defaults need to be None as in other checks (try `check.to_dict` with these defaults)
- please add `number_rows/fields` as it's asked here - https://github.com/frictionlessdata/frictionless-py/issues/820#issuecomment-825526182 (it's basically a shorthand for setting min/max to the same number)

frictionless/checks/regulation.py
```diff
@@ -59,6 +59,80 @@ def validate_row(self, row):
...
+            )
+        if number_fields > self.__max_fields:
+            yield errors.TableDimensionsError(
+                note="Current number of fields is %s, the maximum is %s"
+                % (number_fields, self.__max_fields)
+            )
+
+    def validate_row(self, row):
+        self.__last_row = row
+        if self.__max_rows > 0 and self.__last_row.row_number > self.__max_rows:
```
roll, line 139: - it might be better to move this condition to `validate_end` so it's run only once not for every row

frictionless/checks/regulation.py
```diff
@@ -59,6 +59,80 @@ def validate_row(self, row):
     }
 
 
+class table_dimensions(Check):
```
roll, line 62: - please add tests

roll (changes requested):
Thanks @aivuk, I've added some comments

frictionless/checks/regulation.py
aivuk: How to specify that the required must be one of a set?

roll: I'm not sure but I think there should be a way to do so with JSON Schema. If not we can validate it programatically

aivuk: I'm using OneOf as described here https://json-schema.org/understanding-json-schema/reference/combining.html#oneof

frictionless/checks/regulation.py
aivuk: Changed to None

roll:
Hi, @aivuk, It already looks good in my opinion. Just in case I suggest adding a few edge case tests like:
- more than one limit provided
- no limits provided
- limits provided inside a descriptor (not as an arg)

aivuk:
I added other tests, but what do you mean with:

> limits provided inside a descriptor (not as an arg)

Creating a schema? Do you have an example?

aivuk:
Added the declarative tests @roll

tests/checks/test_regulation.py
```diff
@@ -164,3 +164,215 @@ def test_validate_row_constraint_list_in_formula_issue_817():
...
+        "data/table-limits.csv",
+        checks=[checks.regulation.table_dimensions(min_fields=4, max_rows=3)],
+    )
+    assert report.flatten(["limits", "code"]) == []
+
+
+def test_validate_table_dimensions_min_fields_max_rows_correct_declarative():
+    report = validate(
+        "data/table-limits.csv",
+        checks=[{"code": "table-dimensions", "min_fields": 4, "max_rows": 3}],
```
roll: @aivuk 
We need to use camelCase for property naming inside declarative descriptors (I think I mentioned it somewhere previously) i.e. "minFields", "maxRows" (and others):
- https://framework.frictionlessdata.io/docs/guides/quick-start#usage
- https://github.com/frictionlessdata/frictionless-py/blob/main/frictionless/checks/regulation.py#L52-L59

roll (changes requested):

tests/checks/test_regulation.py
aivuk: I saw it but it was not clear if it was mandatory. I fixed now.

roll (approved):
Thanks @aivuk!

PR #972, opened 2022-01-14
[Tests] don't use error message in assertion

### LINKED ISSUE #971: Tests fail if local set to FR
# Overview

Some tests are using error message in their assertion on top of error code.
But Python can translate those messages depending on locale (at least this is what I experienced on my local french system).
Those tests fail where they should not.

I fix that by removing the error message from assertion tests using only error codes.
A PR is ready to be reviewed, I am creating this issue for reference.

paulgirard:
Those messages can be translated by python depending on locals
Use only error code which is stable
fixes #971


roll (approved):
Thanks, @paulgirard!

I really like the way you chose to resolve the problem (and thanks for finding it) :+1:

PR #981, opened 2022-01-24
Summarize a Package's Resource's

### LINKED ISSUE #980: Create summarized view of a Package's Resources
# Overview

I've been toying around with using frictionless for schema inference on batches of files (a data package in frictionless terminology). While the utility is useful in that it can infer the schema for each file in the batch / resource in the package, it'd help my workflow to be able to view a single, summarized view of all the resources in a package.

I'm imagining functionality that would, for example, look at all the resources and find the _most common_ resource representation of the collection of resources, or identify all the _shared values_ of the resources (i.e. all resources share an encoding and a set of 2 fields).

Ultimately, I think this approach could prove useful for detecting outlier resources in a package - for example, finding the subset of resources that _don't_ share an otherwise uniform schema.

lucas-wave:
- fixes #980

---

This PR implements basic resource summarization, as outlined in #980, using two different "strategies".

For a given `Package`:
- The `most_common` strategy returns a `Resource` that is representative of the majority resource - meaning if `self.resources` contains `Resource`s `[x,x,y,z]`, a view of `Resource` `x` will be returned.
- The `shared_values` strategy returns a `Resource` that is representative of the shared values contained in all `self.resources` - meaning if `self.resources` contains `Resource`s `[{x: 1, y: 2},{x: 1, z: 2}]`, then `Resource` `{x:1}` will be returned.

This functionality could _probably_ be abstracted out of the `Package` class if there are other instances in which "summarizing" multiple instances of the same `Metadata` classes would be useful. I've kept it a part of the `Package` class for now as a proof of concept.


roll:
Closing for now as https://github.com/frictionlessdata/framework/issues/980 had been also closed.

Currently recommended way for `frictionless@5` for a feature like this:
- implementing `package.methods.compare` method

PR #985, opened 2022-01-30
Allows descriptors that subclass collections.abc.Mapping (including dict)

### LINKED ISSUE #906: Consider more flexible field initialization
The `frictionless.Schema` class ignores a list of [anything other than `dict`](https://github.com/frictionlessdata/frictionless-py/blob/bffe82989f0fc782cb51b0c83bf2dee521e55ad8/frictionless/schema.py#L287) for the `fields` keyword. Would you consider accepting any instance of `collections.abc.MutableMapping`? This would not be a breaking change because `isinstance(dict(), collections.abc.MutableMapping)` equals `true`.

My use case: I'm constructing the field metadata in a `pandas.DataFrame` and using the `into` keyword of `pandas.DataFrame.to_dict` to customize the mapping produced. My custom output subclasses `collections.abc.MutableMapping`, but is not an instance of `dict`.

itcarroll:
- fixes #906 

The main change is a couple replacements of `isinstance(descriptor, dict)` with `isinstance(descriptor, collections.abc.Mapping)`. In the added test, I defined a sublcass of collections.abc.Mapping. If you don't like where I put the class statement, happy to move it around.

### DIFF
--- a/frictionless/file.py
+++ b/frictionless/file.py
@@ -1,5 +1,6 @@
 import os
 import glob
+from collections.abc import Mapping
 from pathlib import Path
 from .helpers import cached_property
 from . import settings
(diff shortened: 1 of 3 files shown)


roll (approved):
Thanks! It looks good

PR #998, opened 2022-02-17
Validate example field for Schema fields

### LINKED ISSUE #997: Validate fields[].example for Schema
# Overview

Related to https://github.com/frictionlessdata/specs/pull/753

If a field has an `example` key, make sure that the `example` is valid according to the field's type and constraints.

AntoineAugusti:
- fixes #997 

---

An attempt at validating an optional `example` value for a `Field`. If it is present, make sure it matches types/constraints etc.

### DIFF
--- a/frictionless/field.py
+++ b/frictionless/field.py
@@ -289,12 +289,16 @@ class Field(Metadata):
         return self.get("groupChar", settings.DEFAULT_GROUP_CHAR)
 
     @Metadata.property
+    def has_example(self):
+        return self.example != settings.DEFAULT_EXAMPLE
+
+    @Metadata.property
     def example(self):
         """
         Returns:
             any: example value
         """
-        return self.get("example", None)
+        return self.get("example", settings.DEFAULT_EXAMPLE)
 
     # Expand
 
--- a/frictionless/field.py
+++ b/frictionless/field.py
@@ -58,7 +58,7 @@ class Field(Metadata):
         float_number=None,
         decimal_char=None,
         group_char=None,
-        example=None,
+        example=settings.DEFAULT_EXAMPLE,
         # Extra
         schema=None,
     ):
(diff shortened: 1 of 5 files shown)


AntoineAugusti:
Not sure what I can do in regards to the ` project-assign` CI, let me know

frictionless/schema.py
```diff
@@ -302,6 +302,13 @@ def metadata_validate(self):
             if field.builtin:
                 yield from field.metadata_errors
 
+        # Examples
+        for field in [f for f in self.fields if field.example]:
```
AntoineAugusti, line 306: If you pass a clear `None` in the `example` and you expect a string/integer this will not be checked. Not sure how to approach this, let me know or feel free to edit.

frictionless/field.py
```diff
@@ -286,6 +289,18 @@ def group_char(self):
         """
         return self.get("groupChar", settings.DEFAULT_GROUP_CHAR)
 
+    @Metadata.property
+    def has_example(self):
```
roll: Currently, the metadata system allow us to check the key existence like this `if 'example' in field` so we don't need a special prop for it

frictionless/field.py
```diff
@@ -58,6 +58,7 @@ def __init__(
         float_number=None,
         decimal_char=None,
         group_char=None,
+        example=settings.DEFAULT_EXAMPLE,
```
roll: Let's remove DEFAULT_EXAMPLE as, currently, the codebase style is not to use defaults like this. So it should be not present in the metadata dict if not provided and present if provided

frictionless/field.py
```diff
@@ -77,6 +78,8 @@ def __init__(
         self.setinitial("decimalChar", decimal_char)
         self.setinitial("groupChar", group_char)
         self.setinitial("rdfType", rdf_type)
+        if example != settings.DEFAULT_EXAMPLE:
```
roll: Please see above

frictionless/settings.py
```diff
@@ -83,6 +83,7 @@ def read_asset(*paths, encoding="utf-8"):
     {"type": "year"},
     {"type": "string"},
 ]
+DEFAULT_EXAMPLE = "@@default-dummy-example@@"
```
roll: Let's remove it as mentioned above

roll (changes requested):
Thanks! It looks good! I've added a few comments

AntoineAugusti:
@roll Thanks, I've done the suggested updates

roll (approved):
Thank you!

PR #983, opened 2022-01-25
ForeignKey error: metadata displayed in note

paulgirard:
FK error note are enhanced to help dat curator
- fields from soruce and target table are added
- values from source table
- reference table name
- Tests added from multi field FK validation and note format

Those information are useful when curating from validation errors.
fixes #973

- fixes #<issue-number>

---

Please make sure that all the checks pass. Please add here any additional information regarding this pull request. It's highly recommended that you link this PR to an issue (please create one if it doesn't exist for this PR)

PR #1003, opened 2022-03-11
Enrich foreign key error with more information

### LINKED ISSUE #973: Add ForeignKey fields in ForeignKey Error note
# Overview

When a ForeignKey Error is reported there is no way to know which ForiegnKey didn't validate.
Yet one resource can have multiple ForeignKeys attached to it.
It would help to add the Foreign Key Fields in the Error note.

I made a small edit in my local version of fircitonless-py to add those fields in the error note like this : 
```
{
...
"note": "currency;year;reporting not found in the lookup table as currency;year;reporting",
...
"note": "partner not found in the lookup table as original_name",
...
}
```

Fields are ; separated if multiple and both target and source fields are displayed.

But maybe those should be added as variables in the error objects rather than in the note string?

I can submit a PR if you validate the concept and needs.
It was mandatory in my case to understand what was the issue.

areleu:
- fixes #973

---

Alternative proposal to PR #983.

I did not take over the tests because I think they should be rewritten. I am also not entirelly convinced with the note format but I guess it does the work.


areleu:
@aivuk  can you check this out and tell me if it needs adjustments?

roll:
Thanks @areleu,

I'll review as soon as possible

roll:
Hi @areleu,

I've merged the initial PR to provide a base `error.note` improvement. Please go ahead with this PR (tests/etc) if you still interested in providing extended API for the error.

cc @paulgirard

areleu:
I will pick this up as soon as I can, the tests should not be that complicated afaik.

areleu:
I am picking this up now that v5 is out there. It is looking good, the implementation of ForeignKeyError is a little bit clunky, the main problem is that I could not get the Inheritance from RowError with attrs to work properly. If you have a more elegant solution let me know. I will try to implement some example tests this weekend.

areleu:
I think the feature should be protected with that test. That other account is from my home pc.

frictionless/errors/row.py
```diff
@@ -66,11 +66,72 @@ class PrimaryKeyError(RowError):
...
 
 
-class ForeignKeyError(RowError):
+@attrs.define(kw_only=True)
+class ForeignKeyError(TableError):
     type = "foreign-key"
     title = "ForeignKey Error"
     description = "Values in the foreign key fields should exist in the reference table"
     template = 'Row at position "{rowNumber}" violates the foreign key: {note}'
+    tags = ["#table", "#row"]
```
roll: We don't need to duplicate `tags` and `cells` here in the class definition as it's inherited from the `RowError` class

frictionless/errors/row.py
```diff
@@ -66,11 +66,72 @@ class PrimaryKeyError(RowError):
...
+    cells: List[str]
+    """NOTE: add docs"""
+
+    row_number: int
+    """NOTE: add docs"""
+
+    target_keys: List[str]
+    """NOTE: add docs"""
+
+    source_keys: List[int]
```
roll: List[str]?

frictionless/errors/row.py
```diff
@@ -66,11 +66,72 @@ class PrimaryKeyError(RowError):
...
+    """NOTE: add docs"""
+
+    source_name: List[str]
+    """NOTE: add docs"""
+
+    missing_values: List[int]
```
roll: Sorry not sure what `missing_values` means in this context

frictionless/errors/row.py
```diff
@@ -66,11 +66,72 @@ class PrimaryKeyError(RowError):
...
+    """NOTE: add docs"""
+
+    @classmethod
+    def from_row(
+        cls, row, *, target_keys, source_keys, source_name, missing_values, note
```
roll: Can you please use types in the signature instead of the docstring?

frictionless/errors/row.py
```diff
@@ -66,11 +66,72 @@ class PrimaryKeyError(RowError):
...
+    tags = ["#table", "#row"]
+
+    cells: List[str]
+    """NOTE: add docs"""
+
+    row_number: int
+    """NOTE: add docs"""
+
+    target_keys: List[str]
+    """NOTE: add docs"""
```
roll: Can you please add doc description to `target_keys`, `source_keys`, and `source_name` =)

frictionless/errors/row.py
```diff
@@ -66,11 +66,72 @@ class PrimaryKeyError(RowError):
...
+            source_keys=source_keys,
+            source_name=source_name,
+            missing_values=missing_values,
+        )
+
+    # Metadata
+
+    metadata_profile_patch = {
+        "properties": {
+            "cells": {"type": "array", "items": {"type": "string"}},
```
roll: I think `cells` and `rowNumber` are also inherited from the RowError class (to be checked)

roll (changes requested):
Thanks! It looks good in general

I've added a few comments

frictionless/errors/row.py
areleu: I turned into an Iterable, I have seen applications of the library where the index values are strings and so far it has been working properly.

frictionless/errors/row.py
areleu: I added a description in the docstrings:
"Cell values not found in the lookup table."

areleu:
I solved the problem with the method inheritance, somehow calling the parent method in the child method made it properly work.

Unfortunately I can't properly run pyright locally, it has some node dependencies and I can't use those in my work computer but I suspect the problem with the error compatibility should be already solved.

roll (approved):
Thanks!

roll:
@areleu 
I'm releasing it as `v5.0.0b8` with the naming updated to be aligned with the standards (https://specs.frictionlessdata.io/table-schema/#foreign-keys) - https://framework.frictionlessdata.io/docs/errors/row.html#reference

PR #1019, opened 2022-04-12
Feature/added setters to private properties

### LINKED ISSUE #1014: Make properties accessible eg. for CsvDialect, Detector
# Overview

I am upgrading from Tabulator to Frictionless. I have written a [Frictionless wrapper](https://github.com/OCHA-DAP/hdx-python-utilities/blob/base_downloader/src/hdx/utilities/frictionless_wrapper.py) to try to keep my library's API broadly as it was before.

Properties are not always accessible in Frictionless classes eg. for CsvDialect I must do:
```python
...
dialect = CsvDialect()
delimiter = kwargs.pop("delimiter", None)
if delimiter is not None:
    setattr(dialect, "delimiter", delimiter)
...
```

For Detector, I have to do:
```python
...
detector = kwargs.get("detector", frictionless.Detector())
if infer_types:
    default = None
else:
    default = "string"
field_type = kwargs.pop("field_type", default)
detector._Detector__field_type = field_type
field_float_numbers = kwargs.pop("field_float_numbers", True)
detector._Detector__field_float_numbers = field_float_numbers
...
```

Would it be possible to make all these properties settable after creating an instance of the class?

### LINKED ISSUE #1015: Ability to set Detector properties
# Overview

See https://github.com/frictionlessdata/frictionless-py/issues/1014

Currently, it's not possible to set Detector properties after an instance is created:

```python
detector = Detector(buffer_size=100)
detector.buffer_size = 200 # wont work
```

We want to make it settable probably using an approach we use for other classes in Frictionless (for example, for the Field class):

```python
        if name == "schema":
            self.__schema = value
        else:
            return super().__setattr__(name, value)
```

The list of properties we want to be able to set (all of the current):
- buffer_size
- sample_size
- encoding_function
- encoding_confidence
- field_type
- field_names,
- field_confidence
- field_float_numbers
- field_missing_values
- schema_sync
- schema_patch

# Plan
- [ ] implement property setter using `__setattr__` for Detector 
- [ ] write a few tests for the new functionality in `tests/test_detector.py`
- [ ] create a PR and request a review from @roll

shashigharti:
- fixes #1015

---

Adding getters and setter functions for detector.py class private properties.


shashigharti:
I will create new PR including it so closing it.

Issue #1020, opened 2022-04-12
Document `Detector.encoding_function` parameter


roll:
# Overview

Currently, it's missing in the Detector guide - https://framework.frictionlessdata.io/docs/guides/framework/detector-guide/

roll:
DONE 
https://framework.frictionlessdata.io/docs/guides/framework/detector-guide#encoding-function

PR #1025, opened 2022-04-13
Made private properties public for detector

### LINKED ISSUE #1015: Ability to set Detector properties

shashigharti:
Added getters and setters for private properties
-  Added getters and setters for all private fields
-  Added tests for the above code

Added example for encoding_function (#1020)
- update docs for encoding_function and confidence
- updated missing docstring for encoding_function and confidence

- fixes #1015 #1020
---

### DIFF
--- a/docs/guides/framework/detector-guide.md
+++ b/docs/guides/framework/detector-guide.md
@@ -65,6 +65,24 @@ detector = Detector(sample_size=1000)
 resource = describe("data/country-1.csv", detector=detector)
 ```
 
+
+### Encoding Function
+
+By default, Frictionless encoding_function is None and user can use built in encoding functions. But user has option to implement their own encoding using this feature. The following example simply returns utf-8 encoding but user can add more complex logic to the encoding function.
+
+```python script title="Python"
+from frictionless import Detector
+from frictionless import Resource
+
+detector = Detector(encoding_function=lambda sample: "utf-8")
+with Resource("data/table.csv", detector=detector) as resource:
+  assert resource.encoding == "utf-8"
+  assert resource.sample == [["id", "name"], ["1", "english"], ["2", "中国人"]]
+  assert resource.fragment == [["1", "english"], ["2", "中国人"]]
+  assert resource.header == ["id", "name"]
+```
+
+
 ### Field Type
 
 This option allows manually setting all the field types to a given type. It's useful when you need to skip data casting (setting `any` type) or have everything as a string (setting `string` type):
--- a/frictionless/detector.py
+++ b/frictionless/detector.py
@@ -27,8 +27,12 @@ class Detector:
         buffer_size? (int): The amount of bytes to be extracted as a buffer.
             It defaults to 10000
 
-        sample_size? (int): The amount of rows to be extracted as a samle.
+        sample_size? (int): The amount of rows to be extracted as a sample.
             It defaults to 100
+
+        encoding_function? (func): A custom encoding function for the file.
+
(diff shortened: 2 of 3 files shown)


aivuk (commented):
Your test code need some changes. Is there a reason why you are always doing something like:
```

    detector = Detector(sample_size=1000)
    assert detector.sample_size == 1000
    detector.sample_size = 1000
    assert detector.sample_size == 1000
```
If the this pass ```assert detector.sample_size == 1000``` the next line is redundant ```   detector.sample_size = 1000```. And after this step, why are you testing again for the ``` assert detector.sample_size == 1000```?

aivuk (commented):
In your setter logic you could have used the standard Python setattr method. Instead of a lot of if/else you can have something:

`setattr(self, '_{}'.format(name), value)`

roll:
@shashigharti Yea you might mistakenly have used the same value for testing whether we can set a different value. 

No worries it just needs a quick fix

shashigharti:
Hi @aivuk 

It is because It is testing two aspects one is:

- passing params during initialization
- setting private properties using setters.

I think I am using same values so I will change it to different value for second one.

roll:
> In your setter logic you could have used the standard Python setattr method. Instead of a lot of if/else you can have something:
> 
> `setattr(self, '_{}'.format(name), value)`

@aivuk will it work for private attributes?

shashigharti:
> In your setter logic you could have used the standard Python setattr method. Instead of a lot of if/else you can have something:
> 
> `setattr(self, '_{}'.format(name), value)`

This may not work as I tried it yesterday. So I shifted to if else.

shashigharti:
> @shashigharti Yea you might mistakenly have used the same value for testing whether we can set a different value.
> 
> No worries it just needs a quick fix

Yes. I will fix and push the new changes.

aivuk:
> > In your setter logic you could have used the standard Python setattr method. Instead of a lot of if/else you can have something:
> > `setattr(self, '_{}'.format(name), value)`
> 
> This may not work as I tried it yesterday. So I shifted to if else.

Why it didn't work?

aivuk:
> > > In your setter logic you could have used the standard Python setattr method. Instead of a lot of if/else you can have something:
> > > `setattr(self, '_{}'.format(name), value)`
> > 
> > 
> > This may not work as I tried it yesterday. So I shifted to if else.
> 
> Why it didn't work?

Ok, recursion problem, Got it.

shashigharti:
@aivuk @roll 
Thank you for the suggestion, I have made the changes according to your suggestion.

roll (approved):
Thanks!

PR #1030, opened 2022-04-15
Added property setters and tests

### LINKED ISSUE #1026: Rebase from __setattr__ to property setters
# Overview

In these classes we use `__setattr__` to set writable properties:

```
frictionless/package.py
196:    def __setattr__(self, name, value):
206:            return super().__setattr__(name, value)

frictionless/field.py
93:    def __setattr__(self, name, value):
97:            return super().__setattr__(name, value)

frictionless/detector.py
96:    def __setattr__(self, name, value):
121:            return super().__setattr__(name, value)

frictionless/resource.py
271:    def __setattr__(self, name, value):
283:            return super().__setattr__(name, value)
```
It will be better to use explicit Python property setter with proper documentation. For example:

```python
    @property
    def buffer_size(self):
        """ Add docs """
        return self.__buffer_size

    @buffer_size.setter
    def buffer_size(self, value):
        """ Add docs """
        self.__buffer_size = value
```

### LINKED ISSUE #1027: Small typo in contributing file
I think 'cd portal' is 'cd site' in https://github.com/frictionlessdata/frictionless-py/blob/main/CONTRIBUTING.md

shashigharti:
- fixes #1026 
- fixes #1027

---

- property setters and tests for field class
- property setters and tests detector class
- property setters and tests resource class
- property setters and package class
- fixed typo in contributing doc

---

Looking for feedback. I only applied getters and setters to private properties(those that are not linked with meta parent class). And type hint for 

- 'func' I used 'any' 
- 'Schema' I used 'any' (Did not know if it would be right decision to import Schema class)

And for field value I only tested for initialization. I found out while testing that the value field was not changing using setter.


frictionless/detector.py
```diff
@@ -1,6 +1,8 @@
 import codecs
 import chardet
 from copy import copy, deepcopy
+from typing import List, Dict
+
```
roll: Within the FD codebase we don't usually use a whitespace between import groups

frictionless/detector.py
```diff
@@ -122,48 +124,181 @@ def __setattr__(self, name, value):
```
roll: Shall we remove `__setattr__` method as it's not used anymore?

frictionless/detector.py
```diff
@@ -122,48 +124,181 @@ def __setattr__(self, name, value):
 
     @property
     def buffer_size(self):
```
roll: We can also declare an ouput type here and for other properties using Python Typing

frictionless/detector.py
```diff
@@ -122,48 +124,181 @@ def __setattr__(self, name, value):
...
+        """Returns buffer size of the detector. Default value is 10000.
+        Returns:
+            int: detector buffer size
+        """
         return self.__buffer_size
 
+    @buffer_size.setter
+    def buffer_size(self, value: int):
+        """Sets buffer size for detector.
+        Args:
```
roll: We need to use "Parameters:" instead of "Args:" to make `make docs` work as far as I know

frictionless/field.py
```diff
@@ -92,7 +92,8 @@ def __init__(
 
     def __setattr__(self, name, value):
```
roll: I think this method is not needed anymore

frictionless/package.py
```diff
@@ -330,38 +330,73 @@ def created(self):
         """
         return self.get("created", "")
```
roll: Remove `__setattr__` method as it's not implemented using python property setters?

frictionless/resource.py
```diff
@@ -576,13 +576,21 @@ def header(self):
         """
         return self.__header
```
roll: Remove __setattr__ method as it's not implemented using python property setters?

frictionless/package.py
```diff
@@ -330,38 +330,73 @@ def created(self):
...
+
+    @property
+    def onerror(self):
+        """Returns the onerror behaviour/action of the package. Default value is 'ignore'.
         Returns:
-            ignore|warn|raise: on error bahaviour
+            str: package on error behaviour/action(ignore|warn|raise)
         """
-        return self.__onerror
+        return self.__onerror or settings.DEFAULT_ONERROR
```
roll: I think we don't need to use the default here as it's set in `Package.__init__`

tests/test_detector.py
```diff
@@ -130,8 +132,8 @@ def test_detector_set_field_float_numbers():
 def test_detector_set_field_missing_values():
     detector = Detector(field_missing_values=["", "67"])
     assert detector.field_missing_values == ["", "67"]
-    detector.field_missing_values = ["", "70"]
```
roll: I think it's better to keep any existent test assertions to be sure that we don't introduce any breaking changes

tests/test_field.py
```diff
@@ -335,3 +335,26 @@ def test_field_to_copy():
     target = source.to_copy()
     assert source is not target
     assert source == target
+
+
+def test_package_set_schema():
```
roll: test_field?

tests/test_package.py
```diff
@@ -1071,3 +1071,35 @@ def test_package_validation_duplicate_resource_names_issue_942():
     errors = package.metadata_errors
     assert len(errors) == 1
     assert errors[0].note == "names of the resources are not unique"
+
+
+def test_package_set_hashing():
+    package = Package(hashing="SHA-1")
+    assert package.hashing == "SHA-1"
+
```
roll: Let's try not to use empty lines here too. It's not worse or better just let's be consistent with the codebase style

tests/test_resource.py
```diff
@@ -2709,3 +2709,47 @@ def test_resource_preserve_format_from_descriptor_on_infer_issue_188():
                 "rows": 3,
             },
         }
+
+
+def test_package_set_base_path():
```
roll: test_resource?

roll (changes requested):
Thanks @shashigharti!

Looks good in-general, I've added a few suggestions

roll (approved):
Great!

frictionless/detector.py
shashigharti: @roll I have removed it.

shashigharti:
@roll Thanks for the feedback. I have made the changes as per your feedback.

PR #1032, opened 2022-04-15
[WIP] Fixes bug #1005, removing dots from a number when the datapackage

### LINKED ISSUE #1005: Incompatible decimalChar in table schema and csv data files
# Overview

If a field has a period (`.`) as a decimal char in the table schema but the actual csv file uses comma (`,`), a `type-error` is reported in `frictionless validate`. 

However, if the situation is reversed (comma in table schema but period in csv) no error is reported and `1.234` will be interpreted as `Decimal('1.234')`, which might not be the case in the actual data (eg. because of a missing `groupChar`).

Shouldn't an error be reported in both cases? Here is an reproducible example for the schema

```json
"fields": [
  {
    "name": "PERIOD_DECIMAL_CHAR",
    "type": "number",
    "decimalChar": "."
  },
  {
    "name": "COMMA_DECIMAL_CHAR",
    "type": "number",
    "decimalChar": ","
  }
]
```

and data

```
PERIOD_DECIMAL_CHAR;COMMA_DECIMAL_CHAR
5,678;1.234
```

```python
>>> from frictionless import Package, validate_resource
>>> dp = Package('https://raw.githubusercontent.com/dados-mg/datapackage-reprex/invalid-decimal-char/datapackage.json')
>>> resource = dp.get_resource('resource')
>>> report = validate_resource(resource)
>>> rows = resource.read_rows()
>>> report.flatten(["rowPosition", "fieldPosition", "code", "message"])
[[2, 
1, 
'type-error', 
'Type error in the cell "5,678" in row "2" and field "PERIOD_DECIMAL_CHAR" at position "1": type is "number/default"']]
>>> rows[0]
{'PERIOD_DECIMAL_CHAR': None, 'COMMA_DECIMAL_CHAR': Decimal('1.234')}
```

aivuk:
If a decimal character separator is specified in a datapackage, test if the number has a ".", in that case replace the dot for a "|" and this will make the number parsing to fail.

Fixes #1005


roll (changes requested):
Thanks, I agree that it might be too haskiesh.

I added my idea here - https://github.com/frictionlessdata/frictionless-py/issues/1005#issuecomment-1101194467

Can you please start from writing tests for all the edge cases including "bareNumber" and then we will find a proper fix logic

roll:
CLOSED in favor of #1089

PR #1040, opened 2022-04-20
added function to export table data to excel template

### LINKED ISSUE #584: Integrate "Table Schema -> Excel Templates" library
# Overview

There is a great library - https://github.com/hubmapconsortium/tableschema-to-template - to create Excel Templates from Table Schema. We can provide a convenient wrapper e.g. `schema.to_excel_template()`.

shashigharti:
* added function to export table data to excel template.
* added test to test the file creation.
* I used the code from https://github.com/hubmapconsortium/tableschema-to-template/blob/main/tests/test_create_xlsx.py to test xls file creation and its content.
*   looking for feedback.

- fixes #584


frictionless/schema.py
```diff
@@ -270,6 +270,20 @@ def from_jsonschema(profile):
 
         return schema
 
+    def to_excel_template(self, xlsx_path: str) -> any:
```
roll: I would use just `path` to be consistent with other methods

frictionless/schema.py
```diff
@@ -270,6 +270,20 @@ def from_jsonschema(profile):
...
+        """Export schema data to excel
```
roll: I would not use word `data` here as a schema is a metadata. Just `export schema as an excel template` I think

setup.py
```diff
@@ -39,6 +39,7 @@ def read(*paths):
     "pytest-timeout",
     "pydoc-markdown",
     "docstring-parser",
+    "yattag"
```
roll: We need to put all the new dependencies into the excel section of the EXTRAS_REQUIRE. And then using `helpers.import_from_plugin` will make it show a nice error is it's not installed as `fricitonless['excel']`

roll (commented):
Thanks. I think it's a good approach

shashigharti:
Thank you for the feedback! @roll. I will make the changes and update the PR.

shashigharti:
@roll I have made changes as per your feedback.

setup.py
```diff
@@ -38,12 +38,12 @@ def read(*paths):
...
     "pytest-timeout",
     "pydoc-markdown",
-    "docstring-parser",
+    "docstring-parser"    
 ]
 EXTRAS_REQUIRE = {
     "bigquery": ["google-api-python-client>=1.12.1"],
     "ckan": ["ckanapi>=4.3"],
-    "excel": ["openpyxl>=3.0", "xlrd>=1.2", "xlwt>=1.2"],
+    "excel": ["openpyxl>=3.0", "xlrd>=1.2", "xlwt>=1.2", "yattag>=1.14.0"],
```
aivuk: We do need to add "tableschema-to-template>=0.0.12" as a dependency

aivuk (changes requested):
Except by the tableschema-to-template dependency in setup.py, the PR looks good, good job!

tests/test_schema.py
```diff
@@ -469,6 +473,47 @@ def test_schema_not_supported_type_issue_goodatbles_304():
     assert schema.fields[1] == {"name": "age", "type": "bad"}
 
 
+unzipped_dir = "tests/fixtures/output-unzipped"
+
+
+@pytest.mark.parametrize(
+    "zip_path",
+    [
+        str((Path(base) / file).relative_to(unzipped_dir))
```
roll: @shashigharti 
I meant rewriting this line to somethin like:
Path('/'.join(Path(base), file)).relative_to

As ZIP uses "/" even on windows

shashigharti: sure I will change that!. I forgot to update that part.

Issue #891, opened 2021-08-02
Improve type detection algorithm for spare data


roll:
# Overview

See the discussion in the chat:

- https://discordapp.com/invite/Sewv6av
- https://discord.com/channels/695635777199145130/695635777199145133/871416382305239040
- https://discord.com/channels/695635777199145130/695635777199145133/871730478271574047

roll:
As discussed on the sprint planning - this probelm has been resolved by https://github.com/frictionlessdata/frictionless-py/pull/1051. We only need to add synthetic tests to close this issue

PR #1051, opened 2022-04-26
Type Detection - Evaluate scores after a missing value

### LINKED ISSUE #1050: Type detection is incorrect for sparse dataset
With a sparse dataset, the type is not detected correctly. It is easiest to show by an example which I expect should be detected as integer, but in fact resolves to 'any'. 

```
import frictionless

with frictionless.Resource(
        data=[["Heading1"], ['0'], ['1'], [''], [''], ['']],
        trusted=True,
        detector=frictionless.Detector(
            sample_size=6,
            field_confidence=0.5
        )
) as resource:
    print(resource.schema)

>>> {'fields': [{'name': 'Heading1', 'type': 'any'}]}
```
100% of the populated values are integer, the problem is that the scoring is not re-evaluated when a missing value is encountered. I have a code change which fixes the issue. I will put it in a PR.

rad-pat:
- fixes https://github.com/frictionlessdata/frictionless-py/issues/1050

Adds a test for the sparse dataset type detection

### DIFF
--- a/frictionless/detector.py
+++ b/frictionless/detector.py
@@ -446,21 +446,26 @@ class Detector:
             # Infer fields
             fields = [None] * len(names)
             max_score = [len(fragment)] * len(names)
-            treshold = len(fragment) * (self.__field_confidence - 1)
+            threshold = len(fragment) * (self.__field_confidence - 1)
             for cells in fragment:
                 for index, name in enumerate(names):
                     if fields[index] is not None:
                         continue
                     source = cells[index] if len(cells) > index else None
-                    if source in self.__field_missing_values:
+                    is_field_missing_value = source in self.__field_missing_values
+                    if is_field_missing_value:
                         max_score[index] -= 1
-                        continue
+
                     for runner in runners[index]:
-                        if runner["score"] < treshold:
+                        if runner["score"] < threshold:
                             continue
-                        target, notes = runner["field"].read_cell(source)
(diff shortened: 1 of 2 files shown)


roll:
Thanks @rad-pat!

@aivuk can you please take a look as it seems the same this issue - https://github.com/frictionlessdata/frictionless-py/issues/891

aivuk (approved):
This works well and it fixes #1050. My only comment is that we it would be a good idea to add more extensive tests for sparse data. Maybe generating random data with a subset of them of a specific type, calculating the confidence based of the type and running the type detection on this data. I think we could also check for the extreme case like confidence equals 1.

roll:
I also think that we need to generate some synthetic data for it

rad-pat:
I ran black to format the file. I presume the addition of more tests will be another task rather than specifically for this PR as the test library seems fairly light on extensive tests anyway?

aivuk:
> I ran black to format the file. I presume the addition of more tests will be another task rather than specifically for this PR as the test library seems fairly light on extensive tests anyway?

yes, I agree. Better to merge the fix and let the additional tests as another task. Thanks!

rad-pat:
@roll, @aivuk, is this ok to merge?

roll (approved):

roll:
Thanks @rad-pat!

@aivuk is busy ATM so I'll merge it

PR #1052, opened 2022-04-26
added convert to markdown feature to metadata

### LINKED ISSUE #837: Implement package/resource/schema.to_markdown
# Overview

> https://framework.frictionlessdata.io/docs/guides/describing-data

In Frictionless Framework a user can instantiate a Schema, Resource, and Package objects. All of them are Metadata objects which is a Python dictionary with some additional functionality. 

For example, all these classes have `metadata.to_yaml` that converts the metadata that a package/resource/schema has to a YAML string or file.

Users are interested in having `package/resource/schema.to_markdown` function that will convert metadata to a MD string or file. It might be useful for composing readmes on github etc. See https://github.com/frictionlessdata/project/discussions/665

# Plan

- [ ] Review an [existing external lib](https://github.com/geoffreyaldebert/table-schema-to-markdown) for reference (not sure we can re-use this code)
- [ ] Decide on the way of composing markdown text: using some markdown composer, Jinja template or native Python strings (the latter might be messy) ping @roll to sync
- [ ] Decide on what should be a Markdown output. For example, this one for a Table Schema - https://github.com/geoffreyaldebert/table-schema-to-markdown/blob/master/tests/files/expected_repertoire.md. And this one - https://datahub.io/core/co2-ppm - is a package presentation (of course we can't have interactive parts) ping @roll to sync
- [ ] Prototype all the functions
- [ ] Implement all the functions adding docstrings
- [ ] Write functional tests

shashigharti:
* Added to_markdown feature to Schema, Package and Resource classes
* Added tests for Schema, Package and Resource classes.

- fixes #837 
Looking for feedback on:
- I created a folder for jinja2 templates in frictionless/metadata/templates/markdown.
- For code that I reused, I added "#837" for code credit.
- Overall flow

Thank you!


frictionless/helpers.py
```diff
@@ -4,6 +4,7 @@
 import csv
 import json
 import glob
+import jinja2
```
roll, line 7: I think `jinja2` is not in our `INSTALL_REQUIRES`. We need to add it otherwise it will fail on the user's machine (CI can't catch it)

setup.py
```diff
@@ -49,7 +49,7 @@ def read(*paths):
     "html": ["pyquery>=1.4"],
     "json": ["ijson>=3.0", "jsonlines>=1.2"],
     "ods": ["ezodf>=0.3", "lxml>=4.0"],
-    "pandas": ["pandas>=1.0"],
+    "pandas": ["pandas>=1.0", "tabulate>=0.8.9"],
```
roll: Let's add `tabulate` to `INSTALL_REQUIRES` - it's a common library we will use for other work too

frictionless/helpers.py
```diff
@@ -532,3 +534,103 @@ def __get__(self, instance, owner=None):
                         )
                         raise TypeError(msg) from None
         return val
+
+
+# Markdown
+
+
+# This code was taken from issue#837
```
roll: I would add an URL

frictionless/helpers.py
```diff
@@ -532,3 +534,103 @@ def __get__(self, instance, owner=None):
...
+
+    return jinja2.filters.do_indent(
+        _iter(x, level=0), width=tab * level, first=True, blank=False
+    )
+
+
+def render_markdown(path: str, data: dict) -> str:
+    """Render any JSON-like object as Markdown, using jinja2 template"""
+
+    template_dir = os.path.join(os.path.dirname(__file__), "metadata/templates/markdown")
```
roll: We have a special dir called `assets` to store static assets. So I would move it to `frictionless/assets/tempates`

frictionless/helpers.py
```diff
@@ -532,3 +534,103 @@ def __get__(self, instance, owner=None):
...
+        if level > 0:
+            txt = jinja2.filters.do_indent(txt, width=tab, first=True, blank=False)
+        return txt
+
+    return jinja2.filters.do_indent(
+        _iter(x, level=0), width=tab * level, first=True, blank=False
+    )
+
+
+def render_markdown(path: str, data: dict) -> str:
```
roll: It's just a style thing but I would order the new functions in the helpers file this way:
- render_markdown
- filter_dict
- etc

Why? Because when you read the file for you it's more important to understand the high-level functions first while low-level implementation can be more hidden. I think the best explanations of this and other principles can be found here - https://martinfowler.com/books/refactoring.html

frictionless/metadata.py
```diff
@@ -155,6 +155,26 @@ def to_yaml(self, path=None):
                 raise FrictionlessException(self.__Error(note=str(exc))) from exc
         return text
 
+    def to_markdown(self, path: str = None, table: bool = False) -> str:
+        """Save metadata as a markdown
+
+        Parameters:
+            path (str): target path
```
roll, line 162: `table` parameter is omitted

frictionless/metadata.py
```diff
@@ -155,6 +155,26 @@ def to_yaml(self, path=None):
                 raise FrictionlessException(self.__Error(note=str(exc))) from exc
         return text
 
+    def to_markdown(self, path: str = None, table: bool = False) -> str:
+        """Save metadata as a markdown
```
roll: I think `convert metadata` is better as it doesn't always save it

tests/test_package.py
```diff
@@ -1107,3 +1107,273 @@ def test_package_pprint_1029():
...
+                            "fields": ["main_id"],
+                            "reference": {"resource": "main", "fields": ["id"]},
+                        }
+                    ],
+                },
+            },
+        ],
+    }
+    package = Package(descriptor)
+    expected = """# `package`
```
roll: I would put this text and similar to `tests/fixtures/output-markdown` or something like this. It will be easy to read and maintain

roll (changes requested):
It looks great @shashigharti!

I added some suggestions.

Also, thanks a lot @ezwelty for working on it. Would you like to have a look at this PR. Also, how we can properly credit you?

ezwelty:
@roll  For credit, I guess you could link directly to my specific comments from the function docstrings, and throw in my name/handle if that feels appropriate ('Ethan Welty @ezwelty')?

- https://github.com/frictionlessdata/frictionless-py/issues/837#issuecomment-1108604310 for `json_to_markdown` (and `render_markdown`, depending on the version used)
- https://github.com/frictionlessdata/frictionless-py/issues/837#issuecomment-906246975 for the others

Thanks @shashigharti for implementing this!

Something I notice while skimming the code: In your `render_markdown`, you are pointing `FileSystemLoader` to the built-in template directory, not the parent directory of the file designated by the `path` argument. Therefore, as written, `render_markdown` will not work with an arbitrary template, but only for a template in the template directory. Was this limitation intentional? If so, it may be worth including in the `render_markdown` docstring.

https://github.com/frictionlessdata/frictionless-py/blob/e726b7a8e77902d7570eb5ec00d5eeee9bd7845d/frictionless/helpers.py#L625

```py
def render_markdown(path: str, data: dict) -> str:
    """Render any JSON-like object as Markdown, using jinja2 template"""

    template_dir = os.path.join(os.path.dirname(__file__), "metadata/templates/markdown")
    environ = jinja2.Environment(
        loader=jinja2.FileSystemLoader(template_dir), lstrip_blocks=True, trim_blocks=True
    )
    environ.filters["filter_dict"] = filter_dict
    environ.filters["dict_to_markdown"] = json_to_markdown
    environ.filters["tabulate"] = dicts_to_markdown_table
    template = environ.get_template(path)
    return template.render(**data)
```

Original:

```py
class RelativeEnvironment(jinja2.Environment):
    """Override join_path() to enable relative template paths."""

    def join_path(self, template, parent):
        return os.path.normpath(os.path.join(os.path.dirname(parent), template))

def render_template(path: str, data: dict) -> str:
    dir = os.path.dirname(os.path.abspath(path))
    file = os.path.split(path)[1]
    environment = RelativeEnvironment(
        loader=jinja2.FileSystemLoader(dir), lstrip_blocks=True, trim_blocks=True
    )
    environment.filters["filter_dict"] = filter_dict
    environment.filters["dict_to_markdown"] = json_to_markdown
    environment.filters["tabulate"] = dicts_to_markdown_table
    template = environment.get_template(file)
    return template.render(**data)
```

shashigharti:
@ezwelty thanks again! . I have added the credit information as you have suggested. For path, it was intentional and I was thinking of simple solution without custom template to get started. For now parameter 'path' to_markdown function is the path to save the markdown file. We can add the custom template feature in coming days. @roll  what do you suggest?

shashigharti:
@roll  I have made the changes as per your feedback. It is ready for review. Thank you!

roll (approved):
Great!

roll:
We can create a feature request for supporting custom markdown templates. Although there is a catch - it will have a really massive API as we need to document and protect from breaking changes the context of the template etc. So I would add it as an experimental feature

PR #1056, opened 2022-04-29
899/fixes warnings

### LINKED ISSUE #899: Get rid of openpyxl warning "unknown extension"
# Overview

See the current warning in the thread or running `pytest`. We need to get rig of this warnings: fixing it in the code or ignoring them in pytest

shashigharti:
- Fixes warnings for test_date, test_time, test_datetime
- fixes #899

---

I have fixed other user warnings. But for excel I am thinking of different solutions but need your suggestions on it:

If I disable all userwarnings for openpyxl here([excel.py](http://excel.py/)) in line 215
```
warnings.filterwarnings('ignore', category=UserWarning, module='openpyxl')
```
https://github.com/frictionlessdata/frictionless-py/blob/main/frictionless/plugins/excel.py#L214

One of our tests fail here:
https://github.com/frictionlessdata/frictionless-py/blob/main/tests/test_resource.py#L2654

It is because it is expecting a user warning here. But I think the test is trying to check if it can skip non text cell. So if we change the code to:
```
def test_resource_skip_rows_non_string_cell_issue_320():
    source = "data/issue-320.xlsx"
    dialect = ExcelDialect(fill_merged_cells=True)
    layout = Layout(header_rows=[10, 11, 12])
    # with pytest.warns(UserWarning):
    with Resource(source, dialect=dialect, layout=layout) as resource:
        assert (
            resource.header[7] == "Current Population Analysed % of total county Pop"
        )
```
it works and it also does not change its meaning. And all our tests will pass.

Thank you!

### DIFF
--- a/frictionless/plugins/excel.py
+++ b/frictionless/plugins/excel.py
@@ -5,6 +5,7 @@ import atexit
 import hashlib
 import tempfile
 import datetime
+import warnings
 from itertools import chain
 from ..exception import FrictionlessException
 from ..metadata import Metadata
@@ -211,6 +212,7 @@ class XlsxParser(Parser):
         # To fill merged cells we can't use read-only because
         # `sheet.merged_cell_ranges` is not available in this mode
         try:
+            warnings.filterwarnings("ignore", category=UserWarning, module="openpyxl")
             book = openpyxl.load_workbook(
                 self.loader.byte_stream,
                 read_only=not dialect.fill_merged_cells,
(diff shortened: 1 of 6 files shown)


roll:
@shashigharti 
Sure, let's just drop the warning expectation in the test - it wasn't a "requirement" just a way to hide the warning

shashigharti:
@roll Thank you!. I have made the changes.

roll (approved):
:tada:

PR #1069, opened 2022-05-09
Implemented `checks.deviated_cell`

### LINKED ISSUE #1066: Implement new check for finding cell size outliers
# Overview

Add new feature to detect cell outliners while parsing a file. A valid file with a long quoted cell or any quoteChar spanning multiple rows would cause missing rows. So to handle this we could add a custom check or  an extension for checks.deviated_value

This issue is derived from #1048
CSV file: [issue1048.csv](https://github.com/frictionlessdata/frictionless-py/files/8630087/issue1048.csv)

Example:
When we disable quoting:

```
$ frictionless validate tmp/issue1048.csv --dialect '{"quoteChar": ""}'
# -------
# invalid: tmp/issue1048.csv
# -------

===  =====  ==========  =================================================================
row  field  code        message
===  =====  ==========  =================================================================
 36     11  extra-cell  Row at position "36" has an extra value in field at position "11"
===  =====  ==========  =================================================================
```

shashigharti:
New outlier check
tests for outlier check
typo fixes

- fixes #1066

---
Need feedback:
I tried adding it in deviated_value class but it did not fit into it quite well so I separated it in a different file.


frictionless/checks/cell/outlier_value.py
```diff
@@ -0,0 +1,80 @@
...
+            note = 'outlier value check requires field "%s" to exist'
+            yield errors.CheckError(note=note % self.__field_name)
+        elif self.resource.schema.get_field(self.__field_name).type != "string":
+            note = 'outlier value check requires field "%s" to be string'
+            yield errors.CheckError(note=note % self.__field_name)
+
+    def validate_row(self, row):
+        cell = row[self.__field_name]
+        if cell is not None:
+            self.__cells.append(len(cell))
```
roll: It seems to be not `__cells` but `__cell_sizes`

frictionless/checks/cell/outlier_value.py
```diff
@@ -0,0 +1,80 @@
...
+       descriptor (dict): check's descriptor
+       field_name (str): a field name to check
+       interval? (str): statistical interval (default: 3)
+
+    """
+
+    code = "outlier-value"
+    Errors = [errors.OutlierValueError]
+
+    def __init__(self, descriptor=None, *, field_name=None, interval=None):
```
roll: I think we need to implement this check without asking for a `field_name`:

```python
report = validate( "data/issue-1066.csv", checks=[checks.outlier_value()])
``` 

It needs to check whether string fields (any of them) have values with suspicious size (too long or too short)

frictionless/checks/cell/outlier_value.py
```diff
@@ -0,0 +1,80 @@
+import statistics
+from ... import errors
+from ...check import Check
+
+
+class outlier_value(Check):
```
roll: Maybe `checks.deviated_cell`? (while `deiviated_value` checks values; `deviated_cell` checks cell sizes)

tests/checks/cell/test_outlier_value.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def test_validate_outlier_value_1066():
+    report = validate(
+        "data/issue-1066.csv", checks=[checks.outlier_value(field_name="Gestore")]
+    )
+    assert report.flatten(["code", "note"]) == [
+        [
+            "outlier-value",
+            'value in row at position "35" and field "Gestore" is deviated',
```
roll: I think we need a message like "the cell at ... has deviated size"

roll (changes requested):
Thanks @shashigharti!

I've added my thoughts

shashigharti:
Thank you! @roll. I have made the changes and it is ready for review.

tests/checks/cell/test_deviated_cell.py
```diff
@@ -0,0 +1,27 @@
...
+
+# Issues
+
+
+def test_validate_deviated_cell_1066():
+    report = validate(
+        "data/issue-1066.csv",
+        checks=[
+            checks.deviated_cell(
+                ignore_fields=[
```
roll: What will be the result if we don't ignore these fields?

If it fails without ignoring them we might need to tweek the logic

frictionless/checks/cell/deviated_cell.py
```diff
@@ -0,0 +1,78 @@
+import statistics
+from ... import errors
+from ...check import Check
+from typing import List, Iterator
+
+
+class deviated_cell(Check):
+    """Check if the cell value is deviated
```
roll: cell size

tests/checks/cell/test_deviated_cell.py
shashigharti: If all fields are added it fails. There are many other columns which have deviated value. Sure will check the data and see if they are valid alerts.

roll: I think the reason for it that `deviated_cell` check is just different to `deviated_value` in some aspects. I think for `deviated_cell` we need much bigger default interval to catch really obvious problems as we have in the original CSV.

For example, consider average cell size in column: 50 characters. Even 500 might ok but 5000 is what we need to report. By default it should be really forgiving otherwise we will mark invalid a lot of valid files

shashigharti: yes little bit deviation also triggers the alert. As you have suggested, may be we can use either 5000 or deviated value which ever is greater. In that case we can also remove minimum check as cell size can't be negative.

shashigharti:
Thank you @roll. I have made the changes which checks for threshold value which is either 5000 or maximum value(3 standard deviations from average), which ever is greator. It is ready for review and would like to know your feedback.

roll:
Hi @shashigharti, can you please resolve the conflicts?

Regarding the logic - will it work if some column has really long values like (more than 5000 characters)?

shashigharti:
> Hi @shashigharti, can you please resolve the conflicts?
> 
> Regarding the logic - will it work if some column has really long values like (more than 5000 characters)?

@roll I have tested the logic using synthetic data. Sharing it with you to get your feedback. Once you confirm I will add new tests.
screenshot-docs google com-2022 05 16-18_04_19

shashigharti:
@roll thank you for the guidance!. I have made new changes and added new tests for edge cases (small cell size with/without deviation, large cell size with/without deviation). 

I was also thinking of using IQR method to find the maximum value which is not sensitive to large values. But  3.7 does not support feature quantiles  https://github.com/python/cpython/blob/3.10/Lib/statistics.py#L641

roll (approved):
:tada:

PR #1070, opened 2022-05-09
Added `steps.field_merge` and `steps.field_pack`

### LINKED ISSUE #907: Core functionality missing
# Overview

Hi there! 👋🏻 

While using frictionless we found out, we are missing few core functionalities:
* `field-merge`
* `field-pack`
* `field-rename`
Especially first is cumbersome, as we wanted to create `geometry` column by merging teo columns. Workaround is to `field-add` and use formula to do so, however this is not working in declarative way (that's is our desired way how to have it in our infrastructure).
* `Package`  functionality (write package = write all resources that it has, do something to all Resources, ....)

And one question - when we can expect Storage API?
> The Storage concpet is responsible for reading and writing data package from dataset source like CKAN, SQL, or others. Currently, the Storage API is not yet finished so you can try reading the codebase and implement your own storage but you need to be ready for some changes to the API that might come.

---

shashigharti:
* Added field-merge and field-pack step
* tests for field-merge and field-pack

- fixes #907

---
I could not find any functions in petl so created helper functions for the above functionalities.

### DIFF
--- a/frictionless/steps/field/field_merge.py
+++ b/frictionless/steps/field/field_merge.py
@@ -1,0 +1,51 @@
+from ...step import Step
+from ...field import Field
+from ...helpers import merge
+from typing import List
+
+
+class field_merge(Step):
+    """Merge field"""
+
+    code = "field-merge"
+
+    def __init__(
+        self,
+        descriptor: any = None,
+        *,
+        name: str = None,
+        from_names: List[str] = None,
+        separator: str = "-",
+        preserve: bool = False
+    ):
+        self.setinitial("name", name)
+        self.setinitial("fromNames", from_names)
+        self.setinitial("separator", separator)
+        self.setinitial("preserve", preserve)
+        super().__init__(descriptor)
+
+    # Transform
+
+    def transform_resource(self, resource: any) -> None:
+        table = resource.to_petl()
+        name = self.get("name")
+        from_names = self.get("fromNames")
+        separator = self.get("separator")
+        preserve = self.get("preserve")
+        resource.schema.add_field(Field(name=name))
+        if not preserve:
+            for name in from_names:
+                resource.schema.remove_field(name)
+        resource.data = merge(table, name, from_names, separator, preserve)
+
+    # Metadata
+
+    metadata_profile = {
+        "type": "object",
+        "required": ["name", "fromNames"],
+        "properties": {
+            "name": {"type": "string"},
+            "fromNames": {"type": "array"},
+            "preserve": {},
+        },
+    }
--- a/frictionless/steps/field/field_merge.py
+++ b/frictionless/steps/field/field_merge.py
@@ -1,11 +1,59 @@
 from ...step import Step
 from ...field import Field
-from ...helpers import merge
-from typing import List
+from typing import List, Iterator
+from petl.compat import next, text_type
+
+
+def merge(
+    source: any, name: str, from_names: list, sep: str = "-", preserve: bool = True
+) -> Iterator:
+    it = iter(source)
+
+    hdr = next(it)
+    field_indexes = list()
+    flds = list(map(text_type, hdr))
+
+    # determine output fields
+    outhdr = list(flds)
+    for field in from_names:
+        field_index = flds.index(field)
+        if not preserve:
+            outhdr.remove(field)
+        field_indexes.append(field_index)
+    outhdr.extend([name])
+    yield tuple(outhdr)
+
+    # construct the output data
+    for row in it:
+        value = [v for i, v in enumerate(row) if i in field_indexes]
+        if preserve:
+            out_row = list(row)
+        else:
+            out_row = [v for i, v in enumerate(row) if i not in field_indexes]
+        out_row.extend([sep.join(value)])
+        yield tuple(out_row)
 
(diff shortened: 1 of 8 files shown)


tests/steps/field/test_field_merge.py
```diff
@@ -0,0 +1,55 @@
+from frictionless import Resource, transform, steps
+
+# Issues
```
roll: Sorry it was legacy - we don't need this comment-header (the initial intention was like `bug fixing test`)

tests/steps/field/test_field_merge.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def test_step_field_pack_header_907():
```
roll: merge?

tests/steps/field/test_field_merge.py
```diff
@@ -0,0 +1,55 @@
...
+    target = transform(
+        source,
+        steps=[steps.field_merge(name="details", from_names=["name", "population"])],
+    )
+    assert target.schema == {
+        "fields": [{"name": "id", "type": "integer"}, {"name": "details"}]
+    }
+
+
+def test_step_field_pack_preserve_header_907():
```
roll: I think it would be better to keep metadata/data checks inside one test (the same for `field_pack`)

tests/steps/field/test_field_pack.py
```diff
@@ -0,0 +1,75 @@
...
+                field_type="object",
+                preserve=True,
+            )
+        ],
+    )
+    assert target.read_rows()[0] == {
+        "id": 1,
+        "name": "germany",
+        "population": 83,
+        "details": {"name": "germany", "population": "83"},
```
roll, line 73: We need to pack it as an array by default:
`"details": ["germany", "83"]`

Also, we can implement an object mode of packing also

frictionless/steps/field/field_pack.py
```diff
@@ -0,0 +1,56 @@
...
+        table = resource.to_petl()
+        name = self.get("name")
+        from_names = self.get("fromNames")
+        field_type = self.get("fieldType")
+        preserve = self.get("preserve")
+        resource.schema.add_field(Field(name=name))
+        if not preserve:
+            for name in from_names:
+                resource.schema.remove_field(name)
+        if field_type == "object":
```
roll, line 122: Is it possible to use PETL's functions here instead of copying them into our helpers? (similar for merge)

roll (changes requested):
Thanks @shashigharti!

I added some comments and questions? :+1:

frictionless/steps/field/field_pack.py
shashigharti: PETL doesnot merge and pack function so I wrote it, but I took reference from their code for loops and the flow.

tests/steps/field/test_field_pack.py
shashigharti: I have added both functions array and object mode. If we set data_type to object it packages as json and by default it does package as list ["germany", "83"]

roll: Now I see! Silly me :smile:

frictionless/steps/field/field_pack.py
roll: Wow amazing!

Let's move these functions from `helpers` to corresponding `steps` so related code is close to each other

frictionless/steps/field/field_merge.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # Transform
+
+    def transform_resource(self, resource: any) -> None:
+        table = resource.to_petl()
+        name = self.get("name")
+        from_names = self.get("fromNames")
+        separator = self.get("separator")
+        preserve = self.get("preserve")
+        resource.schema.add_field(Field(name=name))
```
roll: Can we set `type` for a field?

frictionless/steps/field/field_pack.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # Transform
+
+    def transform_resource(self, resource):
+        table = resource.to_petl()
+        name = self.get("name")
+        from_names = self.get("fromNames")
+        field_type = self.get("fieldType")
+        preserve = self.get("preserve")
+        resource.schema.add_field(Field(name=name))
```
roll: The same as for `merge` - let's set a type for the field

shashigharti:
@roll Thank you!. I have made the changes and it is ready for review.

frictionless/steps/field/field_merge.py
```diff
@@ -0,0 +1,104 @@
+from ...step import Step
+from ...field import Field
+from typing import List, Iterator
+from petl.compat import next, text_type
+
+
+def merge(
```
roll, line 7: We need to put low-level functions (internal) at the bottom while having high-level function (API) at the top:
- field_merge
- merge
Similar for the second step. It's better because if someone reads the code he first needs to see high-level functionality etc while low-level implementation might be not so important

roll (approved):
Great work!

roll:
Thanks! I'm merging (btw I've improved the PR title)

PR #1074, opened 2022-05-11
Added true/false_values to detection class.

### LINKED ISSUE #941: New arguments for Detector `field_true/false_values`
# Overview

It's possible to tweak e.g. `missing_values` using Detector but it would be also great to set True/False values without providing a full schema.

shashigharti:
* True false values now can be set from detector class.
* Added tests to test true/false_values feature.

- fixes #941

---

**Feedback required:**
1. Method I am using to set true false values for field.
2. I also tried other ways to pass true_false values in field class which would break previous tests. One of them is:
```
candidate["trueValues"] = self.__field_true_values
candidate["falseValues"] = self.__field_false_values
```

This would break multiple tests because it adds trueValues and falseValues property to fields metadata.


roll:
@shashigharti 
I tried an option similar to what we use for `detector.field_float_numbers` and it seems to be working. WDYT?

roll:
@shashigharti 
I tried an option similar to what we use for `detector.field_float_numbers` and it seems to be working. WDYT?

shashigharti:
> @shashigharti I tried an option similar to what we use for `detector.field_float_numbers` and it seems to be working. WDYT?

@roll Very nice! Thank you it works!. I somehow missed it  :)  . Any other comments on it?

roll (approved):
Thanks! I think it's ready to merge

PR #1085, opened 2022-05-19
reformatted long error messages

### LINKED ISSUE #976: Improve wide error messages formatting in CLI
# Overview

The problem is that some messages go outside of 80-pixel terminal widths and the formatting (lines) looks ugly.

Screenshot from 2022-01-17 17-43-42

Reproducing the problem:
- prepare the dev env - https://framework.frictionlessdata.io/docs/development/contributing
- run `frictionless validate --type resource data/datapackage.json` (it's not a resource so there is a long error message)

shashigharti:
* Reformatted error messages whose length was greator than 90
* Added tests for resource

- fixes #976

---
Feedback required on:

Not quite sure if this method is right way of solving this formatting issue.

- The first half part of the message is not very informative. Real message starts from around 70-80 characters
```
The data resource has an error: "{'format': 'inline',
 'hashing': 'md5',
 'name': 'test-tabulator',
 'profile': 'tabular-data-resource',
 'resources': [{'name': 'first-resource',
                'path': 'table.xls',
                'schema': {'fields': [{'name': 'id', 'type': 'number'},
                                      {'name': 'name', 'type': 'string'}]}},
               {'name': 'number-two',
                'path': 'table-reverse.csv',
                'schema': {'fields': [{'name': 'id', 'type': 'integer'},
                                      {'name': 'name', 'type': 'string'}]}}],
 'scheme': '',
 'stats': {'bytes': 0, 'fields': 0, 'hash': '', 'rows': 0}} is not valid under any of the given schemas
```


aivuk:
@roll I'm looking at it, but I don't think cutting the error message content is a good idea, because an user would like to read the exact message, otherwise how would she know what failed? And in my tests even with the message cut, the "====" used to format the output table headers are being generated bigger than 80 characters.

aivuk:
How about @shashigharti @roll formatting the output with something like https://pypi.org/project/tabulate/ does (or using the library or writing the logic ourselves)? 

```
In [7]: print(tabulate(table, headers=["row", "field", "code", "message"]))
row    field    code            message
-----  -------  --------------  ----------------------------------------------------------------------------------------------------------------------------------------------------
                resource-error  The data resource has an error: "{'fields': [{'name': 'id', 'title': 'Identifier', 'type': 'integer'},
                                            {'name': 'title', 'title': 'Title', 'type': 'string'}],
                                 'foreignKeys': [{'fields': 'id',
                                                  'reference': {'datapackage': 'http://data.okfn.org/data/mydatapackage/',
                                                                'fields': ['id_1', 'title_id'],
                                                                'resource': 'the-resource'}}],
                                 'format': 'inline',
                                 'hashing': 'md5',
                                 'name': 'memory',
                                 'profile': 'tabular-data-resource',
                                 'scheme': '',
                                 'stats': {'bytes': 0, 'fields': 0, 'hash': '', 'rows': 0}} is not valid under any of the given schemas" at "" in metadata and at "oneOf" in profile
```
The only problem that I see is if the user would like to select and copy the error message, but we could add a parameter that makes the cli return the error output in a non tabular format. Maybe something like (or in JSON format):

```
error 1: CODE
field: FIELD
row: ROW
message:
long message here

------------------------------------

error 2: CODE
...
```

shashigharti:
> 

I agree with @aivuk .Using tabulate would also solve the current problem of longer "=====" lines due to long error message.


**Now we are using petl lookall function to display it in tabular format:**
```
petl.util.vis.lookall(
                        [["row", "field", "code", "message"]] + content,
                        vrepr=str,
                        style="simple",
                    )
```

**Using tabulate:**
`tabulate([["row", "field", "code", "message"]] + content)`

Output:
screenshot-nimbusweb me-2022 05 19-15_11_44

shashigharti:
> 

I agree with @aivuk .Using tabulate would also solve the current problem of longer "=====" lines due to long error message.


**Now we are using petl lookall function to display it in tabular format:**
```
petl.util.vis.lookall(
                        [["row", "field", "code", "message"]] + content,
                        vrepr=str,
                        style="simple",
                    )
```

**Using tabulate:**
`tabulate([["row", "field", "code", "message"]] + content)`

Output:
screenshot-nimbusweb me-2022 05 19-15_11_44

aivuk:
@shashigharti I would only suggest to pass ["row", "field", "code", "message"]] as the headers argument to the function tabulate, as I wrote above, because the formatting looks prettier that way :)

shashigharti:
@roll I have made the changes and it is ready for review. Thank you!

@aivuk thank you for your feedback and suggestion.

roll:
Hi @aivuk @shashigharti 

The proposed solution would work well for 1-2 error report but I doubt it would be really useful for 100 errors message.

Is it possible to still stick to one line per error as here:

example

But use some truncation logic on long error message, for example:
- if it's a resource error -- show only "is not valid under..." etc
- if it's a long message - maybe show only error.note
- more smart logic =)

Also, I think having a 2-3 lines error messages also will be fine (consider error 2,4 just has a long message):

example

shashigharti:
@roll thank you!. Shall we also change the error messages with note appended at the end to:

```
'Type error in the cell "{cell}" in row "{rowPosition}" and field "{fieldName}" at position "{fieldPosition}": {note}'
'Row "{rowPosition}" has a type error in field "{fieldName}"'


'The cell "{cell}" in row at position "{rowPosition}" and field "{fieldName}" at position "{fieldPosition}" does not conform to a constraint: {note}'
'Constraint violation in Row "{rowPosition}" at field "{fieldName}"'

or
'Row "{rowPosition}" violates constraint in field "{fieldName}"'
```

shashigharti:
I found that 'tabulate' library has this feature of maxcolwidths but that has not been released:
https://github.com/astanin/python-tabulate/blob/master/tabulate.py#L1242

Current release
https://github.com/astanin/python-tabulate/blob/v0.8.9/tabulate.py#L1216

Here is the conversation for this feature:
https://github.com/astanin/python-tabulate/issues/131

I also tried other libraries which already has this feature 
https://pypi.org/project/texttable/


but tabulate is more simpler and commonly used so I tried adding multiline feature taking reference from tabulate and texttable
@roll @aivuk  WDYT?

frictionless/program/validate.py
```diff
@@ -11,6 +13,31 @@
 from . import common
 
 
+def _wrap_text_to_colwidths(
```
roll: Please put internal code to the bottom of the file so a reader can see more important functions first.
Also, I think we need to add a #NOTE that we're waiting a new tabulate release to remove this func

tests/actions/validate/test_resource.py
```diff
@@ -1199,3 +1199,13 @@ def test_validate_resource_duplicate_labels_with_sync_schema_issue_910():
...
         ],
     ]
+
+
+def test_validate_invalid_resource_long_error_messages():
+    report = validate("data/datapackage.json", type="resource")
+    assert report.flatten(["code", "note"]) == [
+        [
+            "resource-error",
+            "\"{'format': 'inline',\n 'hashing': 'md5',\n 'name': 'test-tabulator',\n 'profile': 'tabular-data-resource',\n 'resources': [{'name': 'first-resource',\n                'path': 'table.xls',\n                'schema': {'fields': [{'name': 'id', 'type': 'number'},\n                                      {'name': 'name', 'type': 'string'}]}},\n               {'name': 'number-two',\n                'path': 'table-reverse.csv',\n                'schema': {'fields': [{'name': 'id', 'type': 'integer'},\n                                      {'name': 'name', 'type': 'string'}]}}],\n 'scheme': '',\n 'stats': {'bytes': 0, 'fields': 0, 'hash': '', 'rows': 0}} is not valid under any of the given schemas\" at \"\" in metadata and at \"oneOf\" in profile",
```
roll: Python validate call shoud not be related to the CLI so not sure about this test

tests/program/test_validate.py
```diff
@@ -219,8 +219,9 @@ def test_program_validate_zipped_resources_979():
     assert result.stdout.count("valid: ogd10_energieforschungstatistik_ch.csv")
     assert result.stdout.count("valid: ogd10_catalogs.zip => finanzquellen.csv")
     assert result.stdout.count("invalid: ogd10_catalogs.zip => capital-invalid.csv")
+    assert result.stdout.count("Schema is not valid")
     assert result.stdout.count(
-        "Schema is not valid: Schemas with duplicate field names are not supported"
+        "Schemas with duplicate field\n                error    names are not supported\n\n"
```
roll: Can we use a multiline string here?

roll (changes requested):
Nice @shashigharti!

Added some comments. Can you also screenshot the result?

frictionless/program/validate.py
roll: I see you added it to the docstring but it's also good to have `#TODO or #NOTE` to make it searchable

tests/program/test_validate.py
shashigharti: I had tried that but it did not work because the stdout in cli stores output string as. 
```
...
Schemas with duplicate field\n                error    names are not supported\n\n
...
```
However this works:
```
    assert result.stdout.count(
        """row    field    code     message
-----  -------  -------  -------------------------------------------------
                schema-  Schema is not valid: Schemas with duplicate field
                error    names are not supported"""
    )
```

shashigharti:
@roll I have made changes as per your feedback and here is the screenshot of the output.

maxwidth

Regarding multiline test I had tried and it had not work. std output looks like this:
```
...
Schemas with duplicate field\n                error    names are not supported\n\n
...
```
So now I used full content that is being displayed in test.

roll:
@shashigharti 
Perfect! Great job!

Please fix tests (I think we need to delete the one I mentioned) and we will be ready to merge

shashigharti:
@roll Thank you!. I have made the changes.

PR #1097, opened 2022-05-24
Added support for TZ data for the Pandas plugin

### LINKED ISSUE #799: Support time zone aware datetimes in the Pandas integration
# Overview
Hello roll and frictionless team. First of all: great work! We are using frictionless in production and it really helps us managing data consistently.
However I noticed that when I read a datapackage or resource with a datetime field that contains time zone aware timestamps, the time zone information is discarded when converting to pandas.

# Reproduce
python version: `3.8.0`
frictionless version: `4.5.1`
pandas version: `1.2.3`

I create a pandas dataframe and write that to a csv file. Infer the csv file with frictionless, read the resource and convert to pandas. Compare the dtypes between the written and read dataframe:

```
import frictionless as fl
import pandas as pd
import pprint

df_out = pd.DataFrame()
df_out["timestamp"] = pd.date_range("2020-01-01", periods=5, freq=pd.to_timedelta("1h"), tz="Europe/Oslo")
df_out["value"] = [3, 5, 7, 9, 4]
df_out.to_csv("data.csv", index=False, date_format="%Y-%m-%dT%H:%M:%S%z")

res = fl.Resource("data.csv")
res.infer()
res.to_json("resource.json")

df_in = fl.Resource("resource.json").to_pandas()

print("DF Write -----------")
print(df_out["timestamp"].dtype)
pprint.pprint(df_out)

print("\nDF Read -----------")
print(df_in["timestamp"].dtype)
pprint.pprint(df_in)
```
Output:
```
DF Write -----------
datetime64[ns, Europe/Oslo]
                  timestamp  value
0 2020-01-01 00:00:00+01:00      3
1 2020-01-01 01:00:00+01:00      5
2 2020-01-01 02:00:00+01:00      7
3 2020-01-01 03:00:00+01:00      9
4 2020-01-01 04:00:00+01:00      4

DF Read -----------
datetime64[ns]
            timestamp  value
0 2020-01-01 00:00:00      3
1 2020-01-01 01:00:00      5
2 2020-01-01 02:00:00      7
3 2020-01-01 03:00:00      9
4 2020-01-01 04:00:00      4
```

This also happens when I manually add `"format": "%Y-%m-%dT%H:%M:%S%z"` to the `resource.json` `timestamp`-field.

Any hints on what I am doing wrong or is it a reproducible bug?

---

aivuk:
- fixes #799

Currently we are convert a tabular resource data to a numpy array and then passing it to pandas.Dataframe constructor. The problem is that the numpy array datetime type has no support to timezone  https://numpy.org/doc/stable/reference/arrays.datetime.html

> Deprecated since version 1.11.0: NumPy does not store timezone information. For backwards compatibility, datetime64 still parses timezone offsets, which it handles by converting to UTC. This behaviour is deprecated and will raise an error in the future.

I removed the convert step to numpy array, passing directly the data_rows to the Dataframe constructor and the timezone information was not lost, but converted to a timezone offset. Even for mixed timezones, the information was there. But now some of the tests for the plugin is failing and I'm looking at why.

### DIFF
--- a/frictionless/plugins/pandas/parser.py
+++ b/frictionless/plugins/pandas/parser.py
@@ -1,5 +1,6 @@
 import isodate
 import datetime
+import decimal
 from ...parser import Parser
 from ...schema import Schema
 from ...field import Field
@@ -42,8 +43,6 @@ class PandasParser(Parser):
                     value = item[field.name]
                 if field.type == "number" and np.isnan(value):
                     value = None
-                elif field.type == "datetime":
-                    value = value.to_pydatetime()
                 cells.append(value)
             yield cells
 
@@ -88,6 +87,8 @@ class PandasParser(Parser):
         if sample is not None:
             if isinstance(sample, (list, tuple)):
                 return "array"
+            elif isinstance(sample, datetime.datetime):
+                return "datetime"
             elif isinstance(sample, datetime.date):
                 return "date"
             elif isinstance(sample, isodate.Duration):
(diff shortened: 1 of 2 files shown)


roll (approved):
Amazing! It now works for write/read round-robin!

PR #1101, opened 2022-05-25
Fix index creation in `to_pandas()` function of resources

### LINKED ISSUE #1100: Resource function to_pandas fails creating index if last field is of type 'datetime'
When reading a resource using `to_pandas()`, index is build from primary key.
But while creating index, code looks into last looked-up field instead of index field to get index data type.
If last field type is of type `datetime` or `date`, index column is tried to be converted to DatetimeIndex, but this fails if primary key column is of type `integer`.

I will provide example package.json and propose a fix in PR.

henhuy:
- fixes #1100 

---

This is an PR to show bug and fix issue #1100


aivuk (approved):
This fix the bug, but testing it using a datetime column as the index, I discovered that it fails when the datetime column contains timezone information. I will create a issue to track this new bug.

aivuk:
And here is the bug created #1105 when I was testing the PR

roll:
Thanks @henhuy @aivuk !

Let's merge this one and continue in #1105

PR #1104, opened 2022-05-26
Add a new compression argument to `to_zip`

### LINKED ISSUE #1103: `to_zip` method of the `Package` class does not really compress the file
# Overview

When using the `.to_zip` method in data packages, the expected output would be a zipped file with the contents of the entire data package, actually compressed with the default zip algorithm.

The file that is actually produced is indeed a zip archive, but it is too large. In fact, it does not appear to be compressed at all as its file size is just over the original uncompressed file's size.

# The test

To test the behavior, just download any zipped csv, create a data package out of it and compare the file sizes.

```python
import os
import urllib.parse
from pathlib import PurePath
import subprocess

import pandas as pd
import requests
from frictionless import Package, Resource

url = 'http://repositorio.dados.gov.br/seges/taxigov/taxigov-corridas-7-dias.zip'
filename = PurePath(urllib.parse.urlparse(url).path).name

# Downloading

response = requests.get(url)
with open(filename, 'wb') as file:
    file.write(response.content)

downloaded_size = os.path.getsize(filename)

# With Pandas

pandas_filename = filename.replace('.zip', '-pandas.zip')
df = pd.read_csv(url, compression='zip')
df.to_csv(pandas_filename, compression='zip')

pandas_size = os.path.getsize(pandas_filename)

# With Frictionless

package = Package()
package.add_resource(Resource(df))
frictionless_filename = filename.replace('.zip', '-frictionless.zip')
package.to_zip(frictionless_filename)

frictionless_size = os.path.getsize(frictionless_filename)

# Unpacked size

subprocess.run(['unzip', filename])
unzipped_filename = filename.replace('.zip', '.csv')
unzipped_size = os.path.getsize(unzipped_filename)

# Size comparison

print(
    f'Downloaded size: {downloaded_size:,}\n'
    f'Pandas size: {pandas_size:,}\n'
    f'Frictionless size: {frictionless_size:,}\n'
    f'Unzipped size: {unzipped_size:,}\n'
)

# Output:
#
# Downloaded size: 219,586
# Pandas size: 225,927
# Frictionless size: 1,119,904
# Unzipped size: 1,118,595
```

# Diagnosis

The problem in `to_zip` seems to be that, [when instancing](/frictionlessdata/frictionless-py/blob/c41f8ef3ccaa83a5d2930e6cda99f64ad919c80e/frictionless/package/package.py#L616) the `ZipFile` class, the `compression` argument is not set.

[As it turns out](https://docs.python.org/3/library/zipfile.html#zipfile-objects), the default value for `compression` is `ZIP_STORED`, which just stores the files without compression. That's why the output is so large.

# Proposed solution

The `to_zip` method should take a `compression` argument as well, but the default value should be different. The reason being that users will expect the file to be actually compressed. Perhaps the default should be `ZIP_DEFLATED`. The value of this argument should then be passed on to the instantiation of `ZipFile`.

augusto-herrmann:
The `to_zip` method of the `Package` class should take a `compression` argument. The default value should be one that actually compresses the file.

- fixes #1103 

It's a pretty simple change. Not sure about the proper way to add a test for this. In issue 1103 I compared the file sizes of a particular zipped csv, but I'm not sure that's how it should be tested.


aivuk (approved):
Good catch @augusto-herrmann, thanks! I tried and it works. I was pretty surprised that the default of the zipfile library is to create a not compressed file, I even tested to confirm.

PR #1108, opened 2022-05-27
Convert to UTC when using a Datetime as an index

### LINKED ISSUE #1105: Exception on pandas parser when using a datetime column containing timezone information as index
If an user try to load a resource and use a datetime column as index, if this column has timezone information the transform to pandas (.to_pandas()) will give an exception:

`ValueError: Tz-aware datetime.datetime cannot be converted to datetime64 unless utc=True`

An example to reproduce this problem:

datapackage.json

```

{
    "name": "to_pandas_bug",
    "resources": [
        {
            "profile": "tabular-data-resource",
            "name": "timezone-issue",
            "path": "timezone-issue.csv",
            "scheme": "file",
            "format": "csv",
            "encoding": "utf-8",
            "schema": {
                "fields": [
                    {"name": "id", "type": "integer"},
                    {
                        "name": "timestamp",
                        "type": "datetime"
                    }
                ],
                "primaryKey": ["timestamp"]
            },
            "dialect": {"delimiter": ";"}
        }
    ]
}
```
And the data timezone-issue.csv:

```
id;timestamp
0;2020-01-01T15:00:00+03:00
1;2020-01-01T15:00:00+04:00
```
@roll

aivuk:
- fixes #1105

Convert a datetime column with timezone information to UTC when this column is being used as an index


roll (approved):
Thanks!

PR #1110, opened 2022-05-28
[WIP] Proof of concept: catching all exceptions immediately

### LINKED ISSUE #840: Implement a debug mode to simplify quick bug analysis
# Overview

We need a debug mode available because errors are caught in a few places in the framework. For example, the CLI catches all the errors or `Report.from_validate`.

We might use an env var or a flag

aivuk:
- fixes #840

Just a proof of concept to see how a catch all exception and raise immediately would work. I added to all "try ; except" blocks a check to see if an environment variable DEBUG was set to 1. In that case I would raise it instead of continue the normal program.

Example of the CLI running without setting the env variable:

```
(fric) ➜  frictionless-py git:(840-adds-debug-mode) ✗ frictionless validate  not-existing-file 
# -------
# invalid: not-existing-file (non-tabular)
# -------

===  =====  ============  ==========================================================================================================
row  field  code          message                                                                                                   
===  =====  ============  ==========================================================================================================
            scheme-error  The data source could not be successfully loaded: [Errno 2] No such file or directory: 'not-existing-file'
===  =====  ============  ==========================================================================================================
```

And with DEBUG=1:

```
(fric) ➜  frictionless-py git:(840-adds-debug-mode) ✗ DEBUG=1 frictionless validate  not-existing-file
Traceback (most recent call last):
  File "/home/aivuk/fric/bin/frictionless", line 33, in <module>
    sys.exit(load_entry_point('frictionless', 'console_scripts', 'frictionless')())
  File "/home/aivuk/fric/lib/python3.10/site-packages/typer-0.4.1-py3.10.egg/typer/main.py", line 214, in __call__
    return get_command(self)(*args, **kwargs)
  File "/home/aivuk/fric/lib/python3.10/site-packages/click-8.1.3-py3.10.egg/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/aivuk/fric/lib/python3.10/site-packages/click-8.1.3-py3.10.egg/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/aivuk/fric/lib/python3.10/site-packages/click-8.1.3-py3.10.egg/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/aivuk/fric/lib/python3.10/site-packages/click-8.1.3-py3.10.egg/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/aivuk/fric/lib/python3.10/site-packages/click-8.1.3-py3.10.egg/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/aivuk/fric/lib/python3.10/site-packages/typer-0.4.1-py3.10.egg/typer/main.py", line 500, in wrapper
    return callback(**use_params)  # type: ignore
  File "/home/aivuk/frictionless/frictionless-py/frictionless/program/validate.py", line 198, in program_validate
    report = validate(source, **options)
  File "/home/aivuk/frictionless/frictionless-py/frictionless/report.py", line 157, in wrapper
    return validate(*args, **kwargs)
  File "/home/aivuk/frictionless/frictionless-py/frictionless/actions/validate.py", line 50, in validate
    return validate(source, deprecate=False, **options)
  File "/home/aivuk/frictionless/frictionless-py/frictionless/report.py", line 157, in wrapper
    return validate(*args, **kwargs)
  File "/home/aivuk/frictionless/frictionless-py/frictionless/actions/validate.py", line 221, in validate_resource
    stats = {key: val for key, val in resource.stats.items() if val}
  File "/home/aivuk/frictionless/frictionless-py/frictionless/helpers.py", line 537, in __get__
    val = self.func(instance)
  File "/home/aivuk/frictionless/frictionless-py/frictionless/resource/resource.py", line 520, in stats
    if self.tabular:
  File "/home/aivuk/frictionless/frictionless-py/frictionless/helpers.py", line 537, in __get__
    val = self.func(instance)
  File "/home/aivuk/frictionless/frictionless-py/frictionless/resource/resource.py", line 649, in tabular
    system.create_parser(self)
  File "/home/aivuk/frictionless/frictionless-py/frictionless/system.py", line 214, in create_parser
    raise FrictionlessException(errors.FormatError(note=note))
frictionless.exception.FrictionlessException: [format-error] The data source could not be successfully parsed: cannot create parser "". Try installing "frictionless-"

```

To add the DEBUG check I used sed in all files with an exception block:
`
find frictionless -name '*py' -exec sed -i 's/^\(\([[:space:]]*\)except .*$\)/\1\n\2    if os.environ.get("DEBUG", 0) == "1":\n\2        raise/g' {} \;`

@roll is that what you had in mind? @shashigharti thoughts?

For me as developer, having the complete stack trace is helpful but as you can see in the example above, without the DEBUG set it's easier to understand the error with the FD standard message, but I believe that in other cases the stack trace will be more helpful, specially if we want to know exactly where to add a breakpoint to debug.

### DIFF
--- a/frictionless/actions/transform.py
+++ b/frictionless/actions/transform.py
@@ -85,6 +85,8 @@ def transform_package(source=None, *, steps, deprecate=True, **options):
         try:
             step.transform_package(package)
         except Exception as exception:
+            if os.environ.get("DEBUG", 0) == "1":
+                raise
             error = errors.StepError(note=f'"{get_name(step)}" raises "{exception}"')
             raise FrictionlessException(error) from exception
 
@@ -163,6 +165,8 @@ def transform_resource(source=None, *, steps, deprecate=True, **options):
         try:
             step.transform_resource(resource)
         except Exception as exception:
+            if os.environ.get("DEBUG", 0) == "1":
+                raise
             error = errors.StepError(note=f'"{get_name(step)}" raises "{exception}"')
             raise FrictionlessException(error) from exception
 
@@ -201,8 +205,13 @@ class DataWithErrorHandling:
         try:
             yield from self.data() if callable(self.data) else self.data
         except Exception as exception:
+            if os.environ.get("DEBUG", 0) == "1":
+                raise
             if isinstance(exception, FrictionlessException):
                 if exception.error.code == "step-error":
                     raise
             error = errors.StepError(note=f'"{get_name(self.step)}" raises "{exception}"')
             raise FrictionlessException(error) from exception
+
+
+import os
--- a/frictionless/actions/validate.py
+++ b/frictionless/actions/validate.py
@@ -119,6 +119,8 @@ def validate_package(
         for resource in package.resources:
             package_stats.append({key: val for key, val in resource.stats.items() if val})
     except FrictionlessException as exception:
+        if os.environ.get("DEBUG", 0) == "1":
+            raise
         return Report(time=timer.time, errors=[exception.error], tasks=[])
 
     # Validate metadata
@@ -219,6 +221,8 @@ def validate_resource(
         stats = {key: val for key, val in resource.stats.items() if val}
         original_resource = resource.to_copy()
     except FrictionlessException as exception:
+        if os.environ.get("DEBUG", 0) == "1":
+            raise
         errors.append(exception.error)
 
     # Open resource
@@ -226,6 +230,8 @@ def validate_resource(
         try:
             resource.open()
         except FrictionlessException as exception:
+            if os.environ.get("DEBUG", 0) == "1":
+                raise
             errors.append(exception.error)
             resource.close()
 
@@ -340,6 +346,8 @@ def validate_schema(source=None, deprecate=True, **options):
         native = isinstance(source, Schema)
         schema = source.to_copy() if native else Schema(source, **options)
     except FrictionlessException as exception:
+        if os.environ.get("DEBUG", 0) == "1":
+            raise
         return Report(time=timer.time, errors=[exception.error], tasks=[])
 
     # Return report
@@ -398,3 +406,6 @@ class ManagedErrors(list):
             if Error.code in self.__scope:
                 continue
             self.__scope.append(Error.code)
+
+
+import os
--- a/frictionless/checks/cell/deviated_cell.py
+++ b/frictionless/checks/cell/deviated_cell.py
@@ -59,6 +59,8 @@ class deviated_cell(Check):
                 average = statistics.median(col_cell_sizes.values())
                 maximum = average + stdev * self.__interval
             except Exception as exception:
+                if os.environ.get("DEBUG", 0) == "1":
+                    raise
                 note = 'calculation issue "%s"' % exception
                 yield errors.DeviatedCellError(note=note)
             # Use threshold or maximum value whichever is higher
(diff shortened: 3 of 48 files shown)


aivuk:
@roll ignore the "import os" at the end of the files (added with another shell command) and don't kill me because of the number of files changed! 😄 For now it's just an experiment.

aivuk:
Something that could be interesting to this feature is if we start to use https://peps.python.org/pep-3134/ to print the full traceback.

Another idea is to follow Flask approach, having a handle_exception function: https://github.com/pallets/flask/blob/a52a7db6c9760adba6ceefa038b6ab69c207d7f9/src/flask/app.py#L1467 and instead of having the check fro the debug flag in all exception code, we create a function to handle the exception and use that in all the framework code.

roll:
Hi @aivuk, it it WIP of ready for a review?

Regarding the implementation, note that we need to implement `system.debug` to mediate debug state - https://github.com/frictionlessdata/frictionless-py/issues/840#issuecomment-1139662334

aivuk:
> Hi @aivuk, it it WIP of ready for a review?
> 
> Regarding the implementation, note that we need to implement `system.debug` to mediate debug state - [#840 (comment)](https://github.com/frictionlessdata/frictionless-py/issues/840#issuecomment-1139662334)

Yes, I saw your comment, getting from the environment variable was just for simplicity of the test above. When we agree on the best way to implement this debug mode I will properly handle the debug flag state.

roll:
Regarding error message vs stacktrace I think we don't need to choose here, we just need to provide an option for the user.

As for myself sometimes I don't understand the bug because of confusing caught error message (like AtrributeError) so I'd like to quickly do something like `frictionless validate bug.csv --debug` to see the stacktrace

aivuk:
> Regarding error message vs stacktrace I think we don't need to choose here, we just need to provide an option for the user.
> 
> As for myself sometimes I don't understand the bug because of confusing caught error message (like AtrributeError) so I'd like to quickly do something like `frictionless validate bug.csv --debug` to see the stacktrace

Yes, but what do you think of the approach in this PR? We would give the user the option to see the full stack trace via --debug, but we would need to add the debug flag check in **all** exceptions like I did. We could create a handle_exception function and start using that and I would refactor the code to use it. 

Another solution would be to use the handle_exception/check the debug flag only in some parts of the program flow (e.g. before creating a report), but in that case I would need your advice to know which parts of the framework architecture it would make sense to add.

roll:
@aivuk 
Yes, majority of the try-except in this PR is a normal flow. So it's not a real user-facing errors in type casting etc. It's just a way to type cast etc. Similar to checks/steps/etc

Real swallowed exceptions are those from:
- validate 
- transform

Ones that go to Report/Status as a text

aivuk:
> @aivuk Yes, majority of the try-except in this PR is a normal flow. So it's not a real user-facing errors in type casting etc. It's just a way to type cast etc. Similar to checks/steps/etc
> 
> Real swallowed exceptions are those from:
> 
>     * validate
> 
>     * transform
> 
> 
> Ones that go to Report/Status as a text

Ok, just confirming if I got this right: do something similar with what I did but getting the flag from the system and only for the validate and transform flow? Because in this case I will close this PR and start a new one.

roll:
:+1:

PR #1111, opened 2022-05-29
Convert datetime to UTC before storing in a SQL DB

### LINKED ISSUE #1102: Support timezone for the SQL plugin
# Overview

Similar to #799 we lose TZ data for the write/read SQL round-robin.

Tests to be updated:
- test_sql_parser_write_timezone_postgresql
- test_sql_parser_write_timezone_mysql
- test_sql_parser_write_timezone_sqlite

aivuk:
- fixes #1102

I did a bit of research on the subject and the recommended way to store datetime in SQL (Postgresql and Sqlite) is to convert it to UTC and if you do want to store the timezone information, you do it in an additional column. What I did in this PR is convert the datetime to UTC, but I'm not adding an additional column with the timezone information. What is missing is just to convert from time to utc time. It's pretty annoying that datetime has a methos astimezone() but the type time hasn't one.


aivuk:
I tested locally only with Sqlite,  this is why the tests failed for MySQL and Postgresql.

aivuk:
> I tested locally only with Sqlite, this is why the tests failed for MySQL and Postgresql.

@roll Do you think that there is a way to prevent that from happening? I ran the test locally with `make test` and there was no fail because the MySQL and Postgresql tests just run if you have set them up. In cases like that, we can only assume that a developer that is working in a feature for plugin X, is going to set all the tests for this plugin before pushing? No automatic way to detect something like: developer worked on code X covered by tests Y,  then on `make test` it would be mandatory to ran tests Y. Just curious if there is a simple solution for these cases. If we had each plugin in an individual repository the answer would be easy.

roll:
I think it's the easiest way to start contributing (when you don't need to setup databases). A contributor will see errors on CI anyway

aivuk:
@roll My last push failed complaining about formatting issues but I did ran `make format` and checked with black, locally there were no errors. Any idea?

roll:
@aivuk 
Maybe you need to upgrade `black` version to the latest? The `make format` changes the resource file for me and I pushed it back

aivuk:
> @aivuk Maybe you need to upgrade `black` version to the latest? The `make format` changes the resource file for me and I pushed it back

I'm already using the latest version `black, 22.3.0 (compiled: yes)`. What is yours?

roll:
The same. It's weird. I'm on py38

aivuk:
I created a new virtual environment with python 3.8, just installed black and tried to format and it didn't do anything also.

roll:
@aivuk 
So what happends now if you pull this branch and run `make format/make lint`?

aivuk:
> make format

It doesn't do anything:

```
(fric38) ➜  frictionless-py git:(1102-support-timezone-sql-plugin) ✗ make lint                      
black frictionless tests --check
All done! ✨ 🍰 ✨
486 files would be left unchanged.
pylama frictionless tests
# mypy frictionless --ignore-missing-imports
(fric38) ➜  frictionless-py git:(1102-support-timezone-sql-plugin) ✗ make format
black frictionless tests
All done! ✨ 🍰 ✨
486 files left unchanged.
(fric38) ➜  frictionless-py git:(1102-support-timezone-sql-plugin) ✗ 
```

roll:
@aivuk 
Sure becase I pushed the fix - https://github.com/frictionlessdata/frictionless-py/pull/1111#issuecomment-1153644290

aivuk:
> @aivuk Sure becase I pushed the fix - [#1111 (comment)](https://github.com/frictionlessdata/frictionless-py/pull/1111#issuecomment-1153644290)

Yes, I saw your fix, but I'm trying to understand what's happening because this happened before (ran black locally, there were no complains, pushed and there were problems in the CI workflow). I'm still working in the commit before your fix to try to debug the problem.

roll:
FINISHED in https://github.com/frictionlessdata/frictionless-py/pull/1197

PR #1112, opened 2022-05-30
Validate single resource

### LINKED ISSUE #221: Validate single resource with `datapackage.json`
Is it possible to validate an individual resource file (CSV) against the schema in `datapackage.json`.

I have separate processing scripts for each resource and like to run `goodtables` for each individually.

Setting `schema` didn't work (as `datapackage.json` is not a Table schema I assume).

shashigharti:
* Added code to handle single resource validation
* Added tests

- fixes #221


frictionless/actions/validate.py
```diff
@@ -103,6 +105,12 @@ def validate_package(
     # Create state
     timer = helpers.Timer()
 
+    # Read resource name for single resource validation
+    resource_name = None
```
roll: I was wondering if we can just use a simpler logic:
- if `resource_name` is provided -> return from `resource.validate` function for this resource

Not sure if it's possible regarding options etc

tests/actions/validate/test_package.py
```diff
@@ -532,3 +532,8 @@ def test_validate_package_with_resource_data_is_a_string_issue_977():
     assert report.flatten() == [
         [None, None, None, "package-error"],
     ]
+
+
+def test_validate_package_single_resource_221():
```
roll, line 536: Let's add more tests (invalid resource/etc). Maybe edge case like wrong resource name

roll (changes requested):
Looks good in-general!

Sorry I didn't write it but we also need to copy the implementation/test to `package.validate` function. Currently, there is a duplication till we release v5

frictionless/actions/validate.py
shashigharti: I have tried revising the code logic:

- The field_name has to be removed from options else it generates error in [line 123](https://github.com/frictionlessdata/frictionless-py/pull/1112/commits/921203d6c461e1991a15f848d17dce74a9ed4132#diff-08a6fec69d71afaa32a62bdda421723363f8235f66a588d3676dc717d639bb9dR123):


 However I could not think of any other solution for the filtering part of resources because the resources is being used in many different places:
- Validating
- Task error generation (parallel and sequential)

So I put it in a different helper function .

tests/actions/validate/test_package.py
shashigharti: I added tests for invalid resource name only. 
`[['task-error', 'The task has an error: No resource found with name ']]`

I could not think of other test case scenarios. We have already handled these:

- CLI  no resource name provided
- Valid resource provided
- Invalid resource name provided

Need some suggestions on it :)

shashigharti:
Thank you! @roll I have made the changes. It is ready for the review.

roll (approved):
Thanks!

roll:
@shashigharti it seems to be failing

Also, can you please add the same tests to `tests/package/validate` tomorrow?

shashigharti:
@roll thank you! I have added new tests and fixed failing tests. It is ready for the review!.

roll (approved):
Thanks!

PR #1115, opened 2022-05-30
Convert a dataframe column to integer when it's in the schema

### LINKED ISSUE #1109: `to_pandas` method of the `Package` class does not preserve integer data type
# Overview

When reading a data package that has a resource with an integer column, which has one or more null values (empty cells in the csv), and then exporting it to Pandas, the integer data type is not preserved and is interpreted as a `float64` instead.

# How to reproduce

Zipping the file may be superfluous, but I wanted to test reading the data and metadata from the same (zipped) file.

```python
import zipfile

import pandas as pd
from frictionless import Package, Resource, describe

# Create the data

data = {
    'id': [1, 2, 3],
    'quantity': [15, None, 20]
}
table = pd.DataFrame(data, dtype='Int64')
assert(table['quantity'].dtype == 'Int64')
table.to_csv('table.csv', index=False)

# Create the package metadata

package = Package(name='zipped-package', title='Example zipped data package')
package.add_resource(describe('table.csv'))
package.to_yaml('datapackage.yaml')

# Pack the zip archive

with zipfile.ZipFile('zipped-package.zip', 'w', compression=zipfile.ZIP_DEFLATED) as archive:
    archive.write('datapackage.yaml')
    archive.write('table.csv')

# Read the zipped package

new_package = Package('zipped-package.zip', descriptor='datapackage.yaml')
new_resource = new_package.get_resource(new_package.resource_names[0])
new_table = new_resource.to_pandas()

df.dtypes
# id            int64
# quantity    float64
# dtype: object

assert(new_table['quantity'].dtype == 'Int64')
```

# Proposal

Maybe the `to_pandas` method should check for a schema. If found, then it would use it to determine the values when calling [read_csv](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) with the appropriate `dtype` argument, passing a dictionary mapping the column names to the appropriate types.

```python
new_table2 = pd.read_csv('table.csv', dtype={'id': 'Int64', 'quantity': 'Int64'})

new_table2.dtypes
# id          Int64
# quantity    Int64
# dtype: object
```

aivuk:
- fixes #1109

---

Pandas plugin sometimes parse a integer column as float if there are empty values. This convert it back to the pandas type Int64 that supports NaN


frictionless/plugins/pandas/parser.py
```diff
@@ -167,6 +167,21 @@ def write_row_stream(self, resource):
 
         # Create/set dataframe
         dataframe = pd.DataFrame(data_rows, index=index, columns=columns)
```
augusto-herrmann, line 169: Perhaps it would be best performance wise to instantiate the DataFrame with the correct `dtype` instead of converting it afterwards.

The problem is that [`pandas.DataFrame`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html#pandas-dataframe) accepts only a single `dtype`. So we would need to do some prep work, like creating a `Series` for each column or something like that.

frictionless/plugins/pandas/parser.py
```diff
@@ -167,6 +167,21 @@ def write_row_stream(self, resource):
...
+        # If there is a float column instead of integer, convert it to the type
+        # Int64 from pandas that supports NaN.
+        # Bug: #1109
+        for field in source.schema.fields:
+            if (
+                field["type"] == "integer"
+                and field["name"] in dataframe.dtypes
+                and str(dataframe.dtypes[field["name"]]) != "int64"
+            ):
+                dataframe[field["name"]] = dataframe[field["name"]].astype("Int64")
```
augusto-herrmann, line 183: This is pretty similar to the workaround I did in my code.

frictionless/plugins/pandas/parser.py
aivuk: Before we used a numpy.array, the problem is that it didn't support datetime with timezone information. Maybe it's better to investigate later if creating separate pandas Series for each column and using them to create the dataframe has a better performance, but for now I would go with the current approach.

roll:
Great! @aivuk @augusto-herrmann so it ok to merge?

frictionless/plugins/pandas/parser.py
```diff
@@ -167,6 +167,21 @@ def write_row_stream(self, resource):
 
         # Create/set dataframe
         dataframe = pd.DataFrame(data_rows, index=index, columns=columns)
+
+        # This step see if there is some column where in the schema is defined
```
augusto-herrmann:
```suggestion
        # This step will see if there is any column for which the schema is defined
```

frictionless/plugins/pandas/parser.py
```diff
@@ -167,6 +167,21 @@ def write_row_stream(self, resource):
...
+        # as 'integer' but Pandas infered as a float. This can happen if there
```
augusto-herrmann:
```suggestion
        # as 'integer' but Pandas inferred it as a float. This can happen if there
```

frictionless/plugins/pandas/parser.py
```diff
@@ -167,6 +167,21 @@ def write_row_stream(self, resource):
...
+        # is a empty value (represented as Not a Number) is the integer column.
```
augusto-herrmann:
```suggestion
        # is an empty value (represented as Not a Number) is the integer column.
```

frictionless/plugins/pandas/parser.py
```diff
@@ -167,6 +167,21 @@ def write_row_stream(self, resource):
...
+        # If there is a float column instead of integer, convert it to the type
```
augusto-herrmann:
```suggestion
        # If the column is of type float instead of integer, convert it to the type
```

frictionless/plugins/pandas/parser.py
```diff
@@ -167,6 +167,21 @@ def write_row_stream(self, resource):
...
+        # Int64 from pandas that supports NaN.
+        # Bug: #1109
+        for field in source.schema.fields:
+            if (
+                field["type"] == "integer"
+                and field["name"] in dataframe.dtypes
```
augusto-herrmann:
```suggestion
                and field["name"] in dataframe.columns
```
Both check works just the same, but checking for presence in columns is more explicit and makes the code easier to understand.

augusto-herrmann (changes requested):
It looks good, @aivuk! I just have some suggestions to improve wording / clarity.

roll (approved):

roll:
Thanks @aivuk @augusto-herrmann !

PR #1117, opened 2022-05-31
Notifies user for invalid fields while validating

### LINKED ISSUE #993: Provide validation hints for common mistakes?
# Overview

For example, one can define `missingValues` on the resource level instead of the schema level. It's a common mistake and according to the standards a resource having `missingValues` is perfectly valid (because props set are customizable and open-ended). In this case (and similar), the framework might emit a warning in the validation report.

---

Missed schema level (`resource -> fields` instead of `resource -> schema -> fields`) - https://github.com/frictionlessdata/frictionless-py/issues/1043

shashigharti:
* Added check for invalid fields: missingValues and fields (in resource descriptor)
* Added check for invalid field: missingValues (in package descriptor)
* Added tests
* Fixed old tests

- fixes #993


shashigharti:
@roll This is ready for the review.

I also tried adding check for the following in package.json. It quietly passes the validation but doesn't trigger metadata_validate function. What would be your suggestion to catch this error?

```
{
  "name": "test-tabulator",
  "fields": [
          {
            "name": "id",
            "type": "integer",            
          },
          {
            "name": "name",
            "type": "string"
          }
        ]
}
```
screenshot-localhost_8888-2022 05 31-17_23_34


Thank you!

frictionless/package/package.py
```diff
@@ -704,6 +704,13 @@ def metadata_process(self):
                 dict.__setitem__(self, "resources", resources)
 
     def metadata_validate(self):
+        # Check invalid properties
+        invalid_fields = {"missingValues": "schema/fields"}
+        for invalid_field, resource in invalid_fields.items():
+            if invalid_field in self:
+                note = f"{invalid_field} should be part of {resource} not package."
+                error = errors.ResourceError(note=note)
+                raise FrictionlessException(error)
```
roll: You need to `yield` errors here as in other places in `metadata_validate` functions (the same for resource)

roll: You can write a test for it:

```
package = Package('with-bad-props')
package.metadata_errors == [...this error...]
```

tests/package/validate/test_general.py
```diff
@@ -345,3 +345,14 @@ def test_validate_package_with_resource_data_is_a_string_issue_977():
...
     ]
+
+
+def test_validate_package_with_missing_values_993():
+    package = Package(descriptor="data/package-with-missingvalues.json")
+    report = package.validate()
+    assert report.flatten(["code", "message"]) == [
+        [
+            "resource-error",
+            "The data resource has an error: missingValues should be part of schema/fields not package.",
```
roll: "missingValues" should be set as "resource.schema.missingValues" (not "resource.missingValues")

tests/resource/validate/test_general.py
```diff
@@ -519,3 +519,25 @@ def test_validate_resource_duplicate_labels_with_sync_schema_issue_910():
...
+    ]
+
+
+def test_validate_resource_fields_without_schema_993():
+    resource = Resource("data/resource-fields-without-schema.json")
+    report = resource.validate()
+    assert report.flatten(["code", "message"]) == [
+        [
+            "resource-error",
+            "The data resource has an error: fields should be part of schema not resource.",
```
roll: "fields" should be set as "resource.schema.fields" (not "resource.fields")

frictionless/package/package.py
```diff
@@ -704,6 +704,13 @@ def metadata_process(self):
                 dict.__setitem__(self, "resources", resources)
 
     def metadata_validate(self):
+        # Check invalid properties
+        invalid_fields = {"missingValues": "schema/fields"}
```
roll: Regarding your question - shall we check for `fields` here also to catch `package.fields`?

roll (changes requested):
Thanks @shashigharti!

I've added a few comments

shashigharti:
Thank you @roll!.  I have made the changes and it is ready for the review.

tests/package/validate/test_general.py
```diff
@@ -345,3 +345,33 @@ def test_validate_package_with_resource_data_is_a_string_issue_977():
...
+        [
+            "resource-error",
+            '"missingValues" should be set as "resource.schema.missingValues" (not "package.missingValues").',
+        ]
+    ]
+
+
+def test_validate_package_with_fields_993():
+    package = Package(descriptor="data/package-with-fields-993.json")
+    output = list(
```
roll: Can we just write assertions like this:
```
assert package.metadata_errors[0].code == ...
assert package.metadata_errors[0].note == ...
```
Your version is smart but really makes the reader think while the simpler one doesn't require any additional energy

tests/package/validate/test_general.py
```diff
@@ -345,3 +345,33 @@ def test_validate_package_with_resource_data_is_a_string_issue_977():
     assert report.flatten() == [
         [None, None, None, "package-error"],
     ]
+
+
+def test_validate_package_with_missing_values_993():
```
roll: I would add also `package.validate` tests

tests/resource/validate/test_general.py
```diff
@@ -519,3 +519,25 @@ def test_validate_resource_duplicate_labels_with_sync_schema_issue_910():
             'Duplicate labels in header is not supported with "schema_sync"',
         ],
     ]
+
+
+def test_validate_resource_with_missing_values_993():
```
roll: I would also add `resource.metadata_errors` tests (so it will be 4 for package and 4 for resource)

roll (changes requested):
@shashigharti Nice!

Just a few minor suggestions

tests/package/validate/test_general.py
shashigharti: Thank you for pointing out. I was making it complex unnecessarily :).

shashigharti:
@roll  Thank you a lot! I have made the changes.

roll (approved):
Perfect!

PR #1130, opened 2022-06-10
1004/adds flag to filter valid invalid rows in extract

### LINKED ISSUE #1004: frictioneless extract flags: to have only the valid lines and only the wrong lines
# Overview

It might be very convenient to have two flags: one to have only the valid lines and one to have only the not valid lines.

Either to debug or to publish only what passes validation.

Thank you

shashigharti:
Added flag(valid/no-valid) to extract command. Changes made:
* Added flag in program -> extract
* Added flag in action -> extract command
* Added flag in package -> extract, resource -> extract
* Added tests for program -> extract
* Added tests for program -> extract -> package, program -> extract -> resource 
* Added tests for resource -> extract, resource -> package
- fixes #1004


roll:
Hi @shashigharti,

Sorry I didn't go trhough the issue before you but I think the request was only related to the CLI:

```bash
$ frictionless extract table/invalid.csv --valid # will extract valid rows
$ frictionless extract table/invalid.csv --invalid # will extract invalid rows
```

Instead of adding the `valid` flag down the road into `extract` functions I would rather add a `filter` parameter (similar to the `process`) making the API more flexible:

```python
valid_rows = extract('data/invalid.csv', filter=lambda row: row.valid)
invalid_rows = extract('data/invalid.csv', filter=lambda row: not row.valid)
# the same to resource/package.extract
```

shashigharti:
@roll  Thank you! I have made the changes and applied filter to cli only like you have suggested. I used the same process the only thing that I changed is the name of the param 'filter' to 'filter_func' because it would clash with predefined 'filter' function.

It is ready for the review!

frictionless/program/extract.py
```diff
@@ -62,6 +62,8 @@ def program_extract(
     yaml: bool = common.yaml,
     json: bool = common.json,
     csv: bool = common.csv,
+    # Row
+    valid: bool = common.valid_rows,
```
roll, line 66: I think it's better to expose also `--invalid` flag (`no-valid` might be confusing). 

Also, in my opinion, using 3 states of the var (True, False and None) is error prone like you use for for `valid=True/False/None`. It's better to have 2 states (True and False/None)

tests/program/test_extract.py
```diff
@@ -200,3 +200,101 @@ def test_program_extract_dialect_keys_option():
...
         path=file, dialect={"keys": ["name", "id"]}
     )
+
+
+def test_program_extract_valid_rows_1004():
+    result = runner.invoke(program, "extract data/countries.csv --valid")
+    assert result.exit_code == 0
+    assert (
+        result.stdout.count("1  Ireland      Britain  67")
+        and result.stdout.count("3  22           Germany  83")
```
roll: This test (and similar) might be really fragile regarding formatting. It will be hard to maintain also. Nothing prevents us to use something like `extract data/countries.csv --valid --csv/json` to get an output in another format. For example, getting a JSON output and using `json.parse` to get a Python dict

frictionless/actions/extract.py
```diff
@@ -6,7 +6,9 @@
 from .. import errors
 
 
-def extract(source=None, *, type=None, process=None, stream=False, **options):
+def extract(
+    source=None, *, type=None, process=None, stream=False, filter_func=None, **options
```
roll: Is it possible to name this arg `filter` instead of `filter_funct`? I think Python will allow it.

We ca use then:

```python
import builtins
data = builtins.filter(filter, data) if filter_func else data
```

roll (changes requested):
@shashigharti 
Looks good! Just a few comments

frictionless/program/extract.py
shashigharti: I think I missed this one in last PR.

shashigharti:
@roll I have made the changes and it is ready for the review!.

roll (approved):
Great!

PR #1137, opened 2022-06-15
1126/fixes package validation error with dialect

### LINKED ISSUE #1126: CLI: frictionless describe fails on --dialect (for --type package)
```
$ frictionless describe --dialect '{"delimiter": ";"}' --type package country-2.csv 
Package.__init__() got an unexpected keyword argument 'dialect'
```
## missing CLI option to describe a package for a directory of CSV files with non-default delimiter
CLI allows to infer schema and resource for a CSV using non-default delimiter such as ";"

However, for package, CLI shows `--dialect` option in help, but using it ends with error message which looks like internal error.

My use case is: having a directory full of CSV files with ";" as a delimiter I would like to describe the whole package using CLI.

shashigharti:
* Added fix for the bug in the package class while using with dialect
* Added tests
- fixes #1126


shashigharti:
@roll it is ready for the review!

frictionless/package/package.py
```diff
@@ -186,6 +190,7 @@ def __init__(
         self.setinitial("name", name)
         self.setinitial("id", id)
         self.setinitial("licenses", licenses)
+        self.setinitial("dialect", dialect)
```
roll: Hi @shashigharti, sorry, in v4 is kinda confusing (will be improved in v5) but can you please store the dialect as `self.__dialect = dialect` the same way as detector (you can see it under the extra comment in the arg list) -- the reason is that dialect is not a part of the data package spec (the same as detector) so we should not store it in the package descriptor

roll (changes requested):

frictionless/package/package.py
roll: The PR looks really well btw :+1:

shashigharti:
Thank you @roll!  It is ready for the review.

roll (approved):
Thanks!

PR #1164, opened 2022-07-07
Fix 1155: fix field_filter field name order in schema + tollerate field name which does not exist

### LINKED ISSUE #1155: field_filter modifies order of labels, but not of values
transformation pipeline provides function `field_filter` which is not documented much.

My initial expectation was it would take the table and provide new one, having only the columns with specified names and the order of fields would not change.

This is definitely not the case. Current behavior is:
- order of field names in records is changed according to order of names in `names` argument
- however, the order of field values in records remains the same as before
- and target.schema remains unchanged in regards to order of field names

## test_field_filter.py
```python
from frictionless import Resource, transform, steps


def test_field_filter():
    rec1 = {"first": "v111", "second": "v222"}

    header = list(rec1.keys())
    assert header == ["first", "second"]
    row1_vals = list(rec1.values())
    assert row1_vals == ["v111", "v222"]

    data = [header, row1_vals]

    resource = Resource(data)
    resource.infer()
    assert resource.labels == header

    assert dict(resource.read_rows()[0]) == rec1
    assert resource.validate()["valid"]
    new_header = ["second", "first"]
    target = transform(
        resource,
        steps=[
            steps.field_filter(names=new_header),
            # steps.table_normalize(),
        ],
    )
    target_row1 = target.read_rows()[0]
    # execting order of headers not modified
    assert target.labels == header
    # expecting records keeping it's values under original name
    assert dict(target_row1) == rec1
```
The test fails:
```
    def test_field_filter():
        rec1 = {"first": "v111", "second": "v222"}

        header = list(rec1.keys())
        assert header == ["first", "second"]
        row1_vals = list(rec1.values())
        assert row1_vals == ["v111", "v222"]

        data = [header, row1_vals]

        resource = Resource(data)
        resource.infer()
        assert resource.labels == header

        assert dict(resource.read_rows()[0]) == rec1
        assert resource.validate()["valid"]
        new_header = ["second", "first"]
        target = transform(
            resource,
            steps=[
                steps.field_filter(names=new_header),
                # steps.table_normalize(),
            ],
        )
        target_row1 = target.read_rows()[0]
        # expecting order of headers not modified
>       assert target.labels == header
E       AssertionError: assert ['second', 'first'] == ['first', 'second']
E         At index 0 diff: 'second' != 'first'
E         Use -v to get more diff
```
## Clarifying expected behavior

I came to two possible behaviors which make sense to me

### var1: pure field filter without changing their order
The `field_filter` shall:
- keep order of field names unchanged
- result shall contain only the names, which are passed by `names` argument
- `target.schema` shall change accordingly
- of course, all field values in all records shall remain bound to original field name

### var2: filter and reorder columns
There are situations where reordering columns (field names + their values) could come handy. Typical situation is when someone provides a table with all required columns but their order is mixed.

I guess this is not the behavior, everyone expects, so I would propose adding a new parameter `reorder` with default value `False`

The new signature would be `target.field_select(name list[str], reorder: bool = False)`

vlcinsky:
- fixes #1155

It fixes field_filter - resulting rows have names and field names properly aligned (order of fields in schema fixed)

The latest commit allows to use field names which are missing.

### DIFF
--- a/frictionless/steps/field/field_filter.py
+++ b/frictionless/steps/field/field_filter.py
@@ -20,9 +20,8 @@ class field_filter(Step):
     def transform_resource(self, resource):
         table = resource.to_petl()
         names = self.get("names")
-        for name in resource.schema.field_names:
-            if name not in names:
-                resource.schema.remove_field(name)
+        schema_fields_dict = {dct["name"]: dct for dct in resource.schema.fields}
+        resource.schema.fields = [schema_fields_dict[name] for name in names]
         resource.data = table.cut(*names)
 
     # Metadata
--- a/frictionless/steps/field/field_filter.py
+++ b/frictionless/steps/field/field_filter.py
@@ -21,8 +21,12 @@ class field_filter(Step):
         table = resource.to_petl()
         names = self.get("names")
         schema_fields_dict = {dct["name"]: dct for dct in resource.schema.fields}
-        resource.schema.fields = [schema_fields_dict[name] for name in names]
-        resource.data = table.cut(*names)
+        new_schema_fields = [
+            schema_fields_dict[name] for name in names if name in schema_fields_dict
+        ]
+        resource.schema.fields = new_schema_fields
+        new_names = [dct["name"] for dct in new_schema_fields]
+        resource.data = table.cut(*new_names)
 
     # Metadata
 
--- a/frictionless/steps/field/field_filter.py
+++ b/frictionless/steps/field/field_filter.py
@@ -19,7 +19,7 @@ class field_filter(Step):
 
     def transform_resource(self, resource):
         table = resource.to_petl()
-        names = self.get("names")
+        names = self.get("names", [])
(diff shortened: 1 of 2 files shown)


vlcinsky:
There are three things topics to evaluate

## Shall we tolerate name, which does not exist in a schema?
The `names` can contain a name, which is not present in given schema.

There are options:
- reject such a request
- try to provide reasonable result omitting columns which do not exist
- provide extra parameter `strict_names` controlling how the code reacts

Provided code is using the 2nd option providing mostly usable results. Anyway, it would fail, if all names would be missing in the schema.

## Shall we fix `primaryKey` or `foreignKey` in the schema?
Modified schema might be missing columns which are still used in `primaryKey` schema property.

I think, it is not easy to decide how to fix the `primaryKey` schema property and we shall leave this to the user who could fix the schema on his own as needed.

The same applies to `foreignKey` property.

# Proposed resolution
Consider existing choice acceptable and approve the PR.

roll (approved):
Thank you!

PR #1179, opened 2022-07-15
1174 add default innerpath

### LINKED ISSUE #1174: Have sensible defaults for `innerpath` in `Package` class
# Overview

Current behaviour when instantiating the `Package` class with a zip file or url is to look for a file named `datapackage.json` inside. If none is found, it throws an exception.

## Example

```python
package = Package('http://repositorio.dados.gov.br/seges/consumo-energia/consumo-energia.zip')
```

this throws an exception because the zip file does not contain a `datapackage.json`, but a `datapackage.yaml`. Specifying an `innerpath` argument makes it work:

```python
package = Package('http://repositorio.dados.gov.br/seges/consumo-energia/consumo-energia.zip', innerpath='datapackage.yaml')
```

## Proposal

In order to make the Frictionless Framework more usable and simplify user's code it should have sensible defaults. It could look first for a `datapackage.json`, then look for a `datapackage.yaml` (maybe also `.yml`), and only if none is found raise an exception.

kagawa0710:
- fixes #1174 

---

When no innerpath argument is given, the existence of data package.json and datapackage.yaml is checked and given.

### DIFF
--- a/frictionless/helpers.py
+++ b/frictionless/helpers.py
@@ -305,15 +305,14 @@ def unzip_descriptor(path, innerpath):
             tempdir = tempfile.mkdtemp()
             zip.extractall(tempdir)
             atexit.register(shutil.rmtree, tempdir)
-            if innerpath != "":
-                pass
-            elif os.path.isfile(os.path.join(tempdir, "datapackage.json")):
+            if innerpath == "":
                 innerpath = "datapackage.json"
-            elif os.path.isfile(os.path.join(tempdir, "datapackage.yaml")):
-                innerpath = "datapackage.yaml"
-            else:
-                # maybe an error
-                innerpath = "datapackage.json"
+                extensions = ("json", "yaml", "yml")
+                default_names = (f"datapackage.{ext}" for ext in extensions)
+                for name in default_names:
+                    if os.path.isfile(os.path.join(tempdir, name)):
+                        innerpath = name
+                        break
             descriptor = os.path.join(tempdir, innerpath)
     return descriptor
 
--- a/frictionless/helpers.py
+++ b/frictionless/helpers.py
@@ -305,6 +305,15 @@ def unzip_descriptor(path, innerpath):
             tempdir = tempfile.mkdtemp()
             zip.extractall(tempdir)
             atexit.register(shutil.rmtree, tempdir)
+            if innerpath != "":
+                pass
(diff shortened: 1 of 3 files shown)


frictionless/helpers.py
```diff
@@ -305,6 +305,15 @@ def unzip_descriptor(path, innerpath):
...
             atexit.register(shutil.rmtree, tempdir)
+            if innerpath != "":
+                pass
+            elif os.path.isfile(os.path.join(tempdir, "datapackage.json")):
+                innerpath = "datapackage.json"
+            elif os.path.isfile(os.path.join(tempdir, "datapackage.yaml")):
+                innerpath = "datapackage.yaml"
+            else:
+                # maybe an error
+                innerpath = "datapackage.json"
```
augusto-herrmann:
```suggestion
            if innerpath == "":
                innerpath = "datapackage.json"
                extensions = ("json", "yaml", "yml")
                default_names = (f"datapackage.{ext}" for ext in extensions)
                for name in default_names:
                    if os.path.isfile(os.path.join(tempdir, name)):
                        innerpath = name
                        break
```

How about we change it like this to avoid some repetition? This also passes the test and has the same effect (plus adds a heuristic for also finding `.yml`).

frictionless/package/package.py
```diff
@@ -99,7 +99,7 @@ class Package(Metadata):
         innerpath? (str): A ZIP datapackage descriptor inner path.
             Path to the package descriptor inside the ZIP datapackage.
             Example: some/folder/datapackage.yaml
-            Default: datapackage.json
+            Default: datapackage.json or datapackage.yaml
```
augusto-herrmann:
```suggestion
            Default: datapackage.json, datapackage.yaml or datapackage.yml
```

augusto-herrmann (changes requested):
Good job, @kagawa0710!

I have run the tests locally and they pass. I suggest a little change to avoid some repetition in code and also catch the case where there is a `datapackage.yml` file.

kagawa0710:
@augusto-herrmann 
Thanks for the suggestion.I think it's better this way too!
Commit accepted.

frictionless/helpers.py
kagawa0710: nice!!

augusto-herrmann (approved):
Excellent. ありがとうございました, @kagawa0710!

roll (approved):
Thanks!

PR #1186, opened 2022-07-17
Support Parquet data format

### LINKED ISSUE #477: Support Parquet data format
# Overview

Parquet integration is a really popular feature request for Frictionless. We want to have integration. At the same time, it's not been discovered yet so this issue requires a design solution proposal. One idea that it can be implemented using `pandas` - https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html

Usually, integration means that we can:

> for resource (internally using `Parser`)

```
# Read
resource = Resource('parquet-file')
resource.read_rows()
# etc

# Write
resource = Resource('table.csv')
resource.write('parquet-file')
```

# Plan
- [ ] research what is Parquet format and how it can be mapped to Frictionless primitives (package/resource/schema) ping @roll to sync
- [ ] TBD

roll:
- fix #477


ewheeler:
Nice! Awesome to see this!

One thing to note-- `fastparquet` and `pyarrow` libraries have some parquet handling differences with pandas:
https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#io-parquet
https://github.com/pandas-dev/pandas/issues/42968#issuecomment-965318185

In the words of a pandas contributor: "Summary: it's a mess :)"

Pandas defaults to `pyarrow` and fails over to `fastparquet`-- but their parquet methods also take an ['engine' param](https://github.com/pandas-dev/pandas/blob/1.4.x/pandas/io/parquet.py#L380) so users can explicitly choose.

Each have their own pros/cons, different dependencies, and different compatibilities with other python packages. For an older version of python (3.4, i think? or maybe it was with a specific numpy version?), it was impossible to have both `pyarrow` and `fastparquet` installed in the same python environment. I'd guess that most folks and projects tend to use one or the other, depending on needs and other dependencies.

Not suggesting one versus the other, but wanted to flag since supporting one engine might limit how `frictionless-py` can be used with existing projects--so worth emphasizing in the docs and/or considering [implementation of an 'engine' param](https://github.com/pandas-dev/pandas/blob/1.4.x/pandas/io/parquet.py#L36) a la pandas.

kindly:
@roll 
Looks good.

My only concern with this is that you are limiting yourself to data that can be loaded "in memory" to a pandas dataframe.  That might be acceptible to begin with.  Howerver, ideally, as parquet is mostly considered a "big data" format and the stated aims of frictionless are:
- Low memory consumption for data of any size
- Reasonable performance on big data

Reading seems not too bad in terms of memroy consumption as you are only reading a single row group (which is essentially a set of rows) at a time, however there is no guarentee the row group will fit in memory.

As for writing I think the whole input is put into a dataframe before writing.  The simplist solution to this is to put many smaller parquet files in a directory as most libararies support treating these files as one large file (including pandas).  So you can chuck up the input into smaller dataframes then write them out to seperate files in that directory.  The other option would be to write some library that takes a stream and uses a lower level parquet writer, but I could not find a decent python one.  

For datapackage_convert I used the rust parquet libarary that could convert from the arrow memory representation in batches.  The arrow library had a CSV reader built in which made it easy, but I am not sure how easy it would be with the dict streams that you use here.

PR #1195, opened 2022-07-18
Added convert command

### LINKED ISSUE #1170: Implement `frictionless convert` command for metadata in CLI
# Overview

We now have some metadata conversion capabilities built into the framework like `to_markdown`, `to_er_diagram`, etc We'd like to expose this functionality to the cli:

```bash
frictionless convert package.json <path> --yaml
frictionless convert package.json <path> --er-diagram
frictionless convert package.json <path> --markdown
# more
```

Note that this action only handle metadata for the data conversion we use `frictinless transform` (at least for now -- later we might think about extending this command to data conversion as a shorcut for the transform)

shashigharti:
Added convert command
Added tests

- fixes #1170

---

### DIFF
--- a/frictionless/program/convert.py
+++ b/frictionless/program/convert.py
@@ -1,0 +1,96 @@
+import typer
+from .main import program
+from ..checklist import Checklist
+from ..dialect import Dialect
+from ..inquiry import Inquiry
+from ..package import Package
+from ..pipeline import Pipeline
+from ..resource import Resource
+from ..report import Report
+from ..schema import Schema
+from ..detector import Detector
+from . import common
+
+
+@program.command(name="convert")
+def program_convert(
+    # Source
+    source: str = common.source,
+    path: str = common.output_file_path,
+    # Command
+    json: bool = common.json,
+    yaml: bool = common.yaml,
+    er_diagram: bool = common.er_diagram,
+    markdown: bool = common.markdown,
+):
+    """Convert metadata to various output"""
+
+    # Validate input
+    if not source:
+        message = 'Providing "source" is required'
+        typer.secho(message, err=True, fg=typer.colors.RED, bold=True)
+        raise typer.Exit(1)
+
+    metadata_type = Detector.detect_descriptor(source)
+    if not metadata_type:
+        message = "File not found"
+        typer.secho(message, err=True, fg=typer.colors.RED, bold=True)
+        raise typer.Exit(1)
+
+    # Initialize metadata
+    try:
+        if metadata_type == "package":
+            metadata = Package.from_descriptor(source)
+        elif metadata_type == "resource":
+            metadata = Resource.from_descriptor(source)
+        elif metadata_type == "schema":
+            metadata = Schema.from_descriptor(source)
+        elif metadata_type == "checklist":
+            metadata = Checklist.from_descriptor(source)
+        elif metadata_type == "dialect":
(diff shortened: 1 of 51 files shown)


shashigharti:
@roll it is ready for review!

frictionless/program/convert.py
```diff
@@ -0,0 +1,53 @@
...
+):
+    """Convert metadata to various output"""
+
+    # Validate input
+    if not source:
+        message = 'Providing package "source" is required'
+        typer.secho(message, err=True, fg=typer.colors.RED, bold=True)
+        raise typer.Exit(1)
+
+    # Initialize Package
```
roll: Sorry @shashigharti,

the initial issue description was quite confusing as a I see now. What I meant that we can have metadata conversion feature actually for all types of metadata (package/resource/schema/etc).

So if we do something like this:

```python
 try:
        metadata_type = Detector.detect_descriptor(source)
        if metadata_type == 'package':
          metadata = Package.from_descriptor(source)
        elif metadata_type == 'resource':
          metadata = Resource.from_description(source)
        # and more types
    except Exception as exception:
        typer.secho(str(exception), err=True, fg=typer.colors.RED, bold=True)
        raise typer.Exit(1)
```

Then we can do:

```python
if yaml:
        content = metadata.to_yaml(path)
        typer.secho(content)
        raise typer.Exit()
```

frictionless/program/convert.py
```diff
@@ -0,0 +1,53 @@
...
+    try:
+        package = Package(source)
+    except Exception as exception:
+        typer.secho(str(exception), err=True, fg=typer.colors.RED, bold=True)
+        raise typer.Exit(1)
+
+    # Return yaml
+    if yaml:
```
roll, line 71: Let's also support `json`. For example, use case:

```
frictionless convert schema.yaml --json
```

frictionless/program/convert.py
```diff
@@ -0,0 +1,53 @@
...
+    # Command
+    yaml: bool = common.yaml,
+    er_diagram: bool = common.er_diagram,
+    markdown: bool = common.markdown,
+):
+    """Convert metadata to various output"""
+
+    # Validate input
+    if not source:
+        message = 'Providing package "source" is required'
```
roll: Providing "source" is required (as it can be other types not only package)

frictionless/program/convert.py
```diff
@@ -0,0 +1,53 @@
...
+        raise typer.Exit(1)
+
+    # Return yaml
+    if yaml:
+        content = package.to_yaml(path)
+        typer.secho(content)
+        raise typer.Exit()
+
+    # Return ER Diagram
+    if er_diagram:
```
roll, line 77: This one will work only for `metadata_type == 'package'`

roll (changes requested):
It's a good start and we can extend it to support all the metadata types

frictionless/program/convert.py
shashigharti: It only detects resource and package only. 
metadata_type = Detector.detect_descriptor(source)

what do you suggest?

shashigharti:
@roll Thank you! I have made the changes and it is ready for the review.

roll (approved):
Great thanks!

roll:
Thanks for adding those convert tests for different metadata types -- it will be really useful to catch regressions :+1:

PR #1209, opened 2022-07-28
1205/recover failing tests

### LINKED ISSUE #1205: Recover specific tests in v5
# Overview

In v5 branch we still have some tests fail because of the codebase changes. For every failing test we need update the test to match a new v5 behaviour (e.g. updating fixture). For some cases, it can be a bug in the codebase (ping @roll to check)

# List of tests

The tests that need to be fixed that way is marked as:

```
@pytest.mark.xfail(reason="issue-1205")
```

We can use search to find them, for example - https://github.com/frictionlessdata/frictionless-py/blob/v5/tests/resource/test_convert.py#L92

shashigharti:
- fixes #1205

Left one tests
- program->test_validate.py
It has error:
seek() takes 2 positional arguments but 3 were given
---


roll (approved):
Thanks!

PR #1226, opened 2022-08-09
Added minimal feature to read/write from/to github repo

### LINKED ISSUE #1185: Integration with Github as a data portal
# Overview

An important step for Frictionless Framework is to provide an ability to read and write packages from different data portals (CKAN/Github/Zenodod/etc) so the users can publish and access their packages easily and using a straightforward API. This issue is for Github integration. The implementation is already prototyped in v5 branch.

### Read package

Read package from a repository that has a datapackage.json/yaml:

```python
package = Package("https://github.com/datasets/population")
package = Package.from_github(...) # alias
```

Read package from a repository without a datapackage.json/yaml. We probably need to filter files and add only CSV/XLS(X) to the package. Also GithubControl should have this configurable. We need to map as much as possible metadata provided for Github Repository:

```python
package = Package("https://github.com/frictionlessdata/repository-demo")
package = Package("<link>", control=portals.GithubControl(formats=['csv']))
```

### Write package

Publish a package on Github (for now, only if the repo doesn't exist). Also we need to provide an ability to store credentials in ENV/etc. We need to map as much as possible metadata provided by Package:

```python
package.to_github(user=, repo=, api_key=)
```

### Read catalog

Read catalog from github search. Design some search configurations like `limit` and `offset` (pagination).

```python
catalog = Catalog(control=portals.GithubControl(search="<frictionless>")
for package in catalog.packages:
  print(package.name)
```

## Plan

- [ ] prototype the functionality based on the functional requirements
- [ ] get feedback from @roll on the implementation
- [ ] finish the implementation
- [ ] design the testing approach (probably using pytest.vcr fro reading but how to test writing?)
- [ ] write a great deal of tests to be sure that the integration works correctly
- [ ] write a comprehensive tutorial - https://framework.frictionlessdata.io/docs/tutorials/tutorials-overview (new section Portals Tutorials)

shashigharti:
Looking for feedback. I have added minimal tests only and will continue to add more tests.

- fixes #1185

---


roll:
Hi @shashigharti, in general I think you made great progress! I'll be commenting in the PR

tests/portals/github/test_manager.py
```diff
@@ -0,0 +1,154 @@
...
+                "path": "https://raw.githubusercontent.com/fdtester/test-repo-without-datapackage/master/data/student.xlsx"
+            },
+        ],
+    }
+
+
+# Write
+
+
+@pytest.mark.vcr
```
roll, line 248: There should be a VCR option to exclude sensitive information from the cassete. See this commit - https://github.com/frictionlessdata/frictionless-py/blob/7db069431b2f5f60c518e5c05dfedafa53504f28/data/cassettes/test_github_parser_write_package_file.yaml - our Github Token is a part of the cassete. Nothing to worry about as Github automatically revoked this token when it hit the public repo but we need a way to exclude it from the cassete

tests/portals/github/test_manager.py
```diff
@@ -0,0 +1,154 @@
...
+import pytest
+from frictionless import portals, FrictionlessException
+from .conftest import options, options_without_dp
+
+
+# Read
+
+
+@pytest.mark.vcr
+def test_github_parser_read(options_without_dp):
```
roll: test_github_manager_read

tests/portals/github/test_manager.py
```diff
@@ -0,0 +1,154 @@
...
+    user = options_without_dp.pop("user")
+    repo = options_without_dp.pop("repo")
+    control = portals.GithubControl()
+    manager = portals.GithubManager(control)
```
roll: In the tests we need to use the Package class instead of the Manager class as per requirements. Manager is an internal class; the framework uses under the hood so the user would write (both should work):

```
package = Package(control=portals.GithubControl(user=user, repo=repo))
package = Package('https://github.com/user/repo")
```

frictionless/portals/github/control.py
```diff
@@ -15,14 +15,38 @@ class GithubControl(Control):
...
+    search: Optional[str] = None
+    """NOTE: add docs"""
+
+    per_page: Optional[int] = 30
+    """NOTE: add docs"""
+
+    page: Optional[int] = None
+    """NOTE: add docs"""
+
+    apikey: Optional[str] = None
```
roll: If you use ENV value here as a default you don't need to handle it anywhere else just use `control.apikey`

```
apikey: Optional[str] = os.environ.get("GITHUB_ACCESS_TOKEN")
```

frictionless/portals/github/manager.py
```diff
@@ -1,45 +1,156 @@
...
-    def read_catalog(self):
-        pass
+    def read_catalog(self, user: Optional[str] = None, apikey: Optional[str] = None):
+        client = platform.github.Github(apikey, per_page=self.control.per_page)
+        user = user or self.control.user
+        search = self.control.search or f"user:{user}"
+        print(search)
+        assert search
+        if self.control.page:
+            search = f"{search}&page={self.control.page}"
```
roll: Maybe using Python's builtin means for building query string. It's not really important but you can try searching for urlib.parse for query strings (can't remember the function)

frictionless/portals/github/control.py
roll: The same for email

frictionless/portals/github/manager.py
```diff
@@ -1,45 +1,156 @@
...
+    while contents:
+        file_content = contents.pop(0)  # type: ignore
+        if file_content.type == "dir":
+            contents.extend(repository.get_contents(file_content.path))  # type: ignore
+        else:
+            paths.append(file_content.path)
+    return paths
+
+
+def _get_package(paths, repository, base_path, formats):
```
roll: It's OK not to use the leading underscore here as it's an internal function anyway. We don't use the leading underscore for cases like this across the codebase

frictionless/portals/github/manager.py
```diff
@@ -1,45 +1,156 @@
...
+                )
+            except Exception as exception:
+                note = "Github API error:" + repr(exception)
+                raise FrictionlessException(note)
+
+        response = namedtuple("response", ["repository", "file_created"])
+        return response(repository, file_created)
+
+
+def _get_contents(contents, repository):
```
roll: I would make this function `get_resources() -> List[Resource]` as on the next step we also need to add some metadata like `resource.name`. Probably the best way for doing so will be:

```
resource = Resource(path=path)
resource.infer(sample=False) # it means infer anything you can without any file reading
```

roll: But if Github API provides metdata like scheme/format/mediatype etc we need to set it directly

```
resource = Resource(path=path, format=format, ...)
```

tests/portals/github/test_manager.py
```diff
@@ -0,0 +1,154 @@
...
+
+# Search
+
+
+@pytest.mark.vcr
+def test_github_parser_read_catalog_with_search():
+    control = portals.GithubControl(search="test user:fdtester")
+    manager = portals.GithubManager(control=control)
+    apikey = os.environ.get("GITHUB_ACCESS_TOKEN", None)
+    packages = manager.read_catalog(apikey=apikey)
```
roll: Note that `manager.read_catalog() -> Catalog` needs to return the Catalog class. So in tests we write:
```
catalog = Catalog(control=control)
for package in catalog.packages:
    ...
```

frictionless/portals/github/control.py
```diff
@@ -15,14 +15,38 @@ class GithubControl(Control):
     user: Optional[str] = None
     """NOTE: add docs"""
 
+    email: Optional[str] = None
+    """NOTE: add docs"""
+
     repo: Optional[str] = None
     """NOTE: add docs"""
 
+    formats: Optional[list] = [".csv"]
```
roll: In our terminology format doesn't contain dots
```
formats: Optional[List[str]] = ['csv']
```

data/fixtures/portals/github/datapackage.json
```diff
@@ -0,0 +1,34 @@
+{
```
roll, line 1: I think we don't need to put the test packates under `fixtures/portals/github` we can just have `data/package-name` and use them across the integrations (ckan/github/zenodo/etc)

roll: I would also have at least 2 resources there

frictionless/portals/github/manager.py
shashigharti: I have removed the query string generation part.

frictionless/portals/github/manager.py
shashigharti: I used resource.infer as I couldn't find any function that would return these file metadata.

get_content is returning all the paths that also includes datapackage.json or datapackage.yaml so was not sure if it could be included as resource get_resources() -> List[Resource]. So I changed the name of the function but return type set as list[str]

data/fixtures/portals/github/datapackage.json
shashigharti: I removed those files and used the existing files datapackage.json.

It has two resources in it.

frictionless/package/package.py
```diff
@@ -441,6 +441,62 @@ def to_ckan(self, target, *, control=None):
...
+        Parameters:
+            source (string): Github repo url e.g. "https://github.com/frictionlessdata/repository-demo"
+            control (portals.GithubControl): Github control
+
+        Returns:
+            Package: package
+        """
+        manager = system.create_manager(source, control=control)
+        if not manager:
+            raise FrictionlessException(f"Not supported Github source: {source}")
```
roll: Maybe also adding control to the error note as source might be None

frictionless/package/package.py
```diff
@@ -441,6 +441,62 @@ def to_ckan(self, target, *, control=None):
         storage.write_package(self, force=True)
         return storage
 
+    @staticmethod
+    def from_github(
```
roll, line 445: What if this function will just use constructor?

I think it will be the same:
```
return Package(source, control=control)
```

WDYT?

frictionless/package/package.py
```diff
@@ -441,6 +441,62 @@ def to_ckan(self, target, *, control=None):
...
+        Returns:
+            Package: package
+        """
+        manager = system.create_manager(source, control=control)
+        if not manager:
+            raise FrictionlessException(f"Not supported Github source: {source}")
+        package = manager.read_package()
+        return package
+
+    def to_github(
```
roll, line 464: What if this function will just call `publish`?

I think it will be the same:
```
return self.publish(source, control=control)
```

roll (commented):
Thanks @shashigharti!

It already looks realy solid I'll continue reviewing when you add resource writing to Github

---

> I added publish function but both to_github and publish has same functionality 'write_package' should I add new function for publish in manager.

I commented in the review regarding this

> Write function only writes datapackage.json file. Shall I add function to write resources?

Yes. We need to publish resource as well. The user story here is that the use will be able upload the whole data package to a github repo 

> I haven't yet added tutorial, I will tutorial once the functionality is completed and approved.

I'm working on the docs improvements now so you will be able to write a tutorials easier than we do now. I'll push changes around tomorrow and we'll discuss

roll:
Hi @aivuk can you please also review/test/participate when you have time?

frictionless/package/package.py
aivuk: I agree with that. Better to keep all the Catalog specific logic inside the portals/github

frictionless/portals/github/manager.py
```diff
@@ -1,45 +1,190 @@
...
-        assert user
-        assert repo
-        location = "/".join([user, repo])
-        repository = client.get_repo(location)
-        branch = repository.default_branch
+    def read_catalog(self) -> Catalog:
+        packages = []
+        query: Dict[str, Any] = {}
+
+        if self.control.repo or self.control.user or self.control.search:
```
roll: I would suggest raising an exception here like:

```
if not self.control.repo and not ...:
    raise
```

It will allow to write the following code without indentation (just slightly easier to read and maintain)

frictionless/portals/github/manager.py
```diff
@@ -1,45 +1,190 @@
...
+                location = "/".join([self.control.user, self.control.repo])
+                try:
+                    repository = client.get_repo(location)
+                except Exception as exception:
+                    note = "Github API error" + repr(exception)
+                    raise FrictionlessException(note)
+                base_path = f"https://raw.githubusercontent.com/{location}/{repository.default_branch}"
+                contents = repository.get_contents("")
+                resource_path = get_resources(contents, repository)
+                packages = get_packages(resource_path, base_path)
```
roll: We can use some data from `control` as a catalog name e.g. a github user name

frictionless/portals/github/manager.py
```diff
@@ -1,45 +1,190 @@
...
+                    branch=repository.default_branch,
+                    committer=author,
+                    author=author,
+                )
+            except Exception as exception:
+                note = "Github API error:" + repr(exception)
+                raise FrictionlessException(note)
+
+        response = namedtuple("response", ["repository", "file_created"])
+        return response(repository, file_created)
```
roll: I think here and in `Packge.to_github` we need to return a full URL to the published data package like: https://raw.githubusercontent.com/fdtester/test-repo-with-datapackage-yaml/master/datapackage.yaml

roll: But I'm not yet sure. WDYT?

tests/portals/github/test_manager.py
```diff
@@ -0,0 +1,404 @@
...
+
+# Read
+
+
+@pytest.mark.vcr
+def test_github_manager_read(options_without_dp):
+    url = options_without_dp.pop("url")
+    output = options_without_dp.pop("output")
+    package = Package(url)
+    assert json.loads(package.to_json()) == output
```
roll: `json.loads(package.to_json())` -> `package.to_descriptor` :wink:

frictionless/portals/github/control.py
```diff
@@ -12,17 +13,57 @@ class GithubControl(Control):
...
     # State
 
-    user: Optional[str] = None
+    apikey: Optional[str] = os.environ.get("GITHUB_ACCESS_TOKEN", None)
+    """NOTE: add docs"""
+
+    email: Optional[str] = os.environ.get("GITHUB_EMAIL", None)
+    """NOTE: add docs"""
+
+    formats: Optional[List[str]] = ["csv"]
```
roll: Let's have ['csv', 'xlsx', 'xls'] by default and also we might have it as DEFAULT_FORMATS const

frictionless/portals/github/control.py
```diff
@@ -12,17 +13,57 @@ class GithubControl(Control):
...
+    name: Optional[str] = os.environ.get("GITHUB_NAME", None)
+    """NOTE: add docs"""
+
+    order: Optional[str] = None
+    """NOTE: add docs"""
+
+    page: Optional[int] = None
+    """NOTE: add docs"""
+
+    per_page: Optional[int] = 30
```
roll: Maybe DEFAULT_PER_PAGE const to the top of the file

tests/portals/github/test_manager.py
```diff
@@ -0,0 +1,404 @@
...
+    package = Package.from_github(url, control=control)
+    assert json.loads(package.to_json()) == {
+        "name": "test-repo-without-datapackage",
+        "resources": [
+            {
+                "name": "student",
+                "type": "table",
+                "path": "https://raw.githubusercontent.com/fdtester/test-repo-without-datapackage/master/data/student.xlsx",
+                "scheme": "https",
+                "format": "xlsx",
```
roll, line 78: It will be great if we can have `encoding` here also. Is it possible to get it from Github API metadata?

tests/portals/github/test_manager.py
```diff
@@ -0,0 +1,404 @@
...
+    catalog = Catalog(repo_url)
+
+    # Read
+    expected_file_path = "data/datapackage.json"
+    with open(expected_file_path) as file:
+        assert json.loads(catalog.packages[0].to_json()) == json.loads(file.read())
+
+
+@pytest.mark.vcr
+def test_github_manager_catalog_from_repo_with_multiple_packages(
```
roll: @shashigharti 
Let's keep it simple: one repo == one data package so we don't need to create a catalog from one repo

roll:
@shashigharti 
Another test suggestion:
```
catalog = Catalog("https://github.com/fdtester")
print(catalog)
```
Currently, it fails for me I guess because we need to de-duplicate package names. I think we might run into the same problem for resource if we have a repo like this:
```
data/table.csv
table.csv
```

BTW can we also have a test repo like this (with multiple files without a data package):

```
legacy/table.xls
table.csv
table.xlsx
```

frictionless/portals/github/manager.py
shashigharti: I think returning the package file makes sense. Or may be returning the Repository only and we can use this method of `Repository` to test the package file. 
`repository.get_contents(package_filename)`

shashigharti:
@roll @aivuk Thank you very much for the feedback!. I have made the changes as per your feedback and it is ready for the review.

I also made some changes to the repo because of duplicate resources and changed previous failing test.

roll:
Thanks @shashigharti! I'll ve rewewing it today or tomorrow

shashigharti:
Closing this PR and creating a new one.

PR #1228, opened 2022-08-12
[ Issue#1227 ] Add itterrows method to Resource

### LINKED ISSUE #1227: Add iterrows method to Resource to avoid replacing the default __iter__ method
# Overview

I think this is critical, I was trying to use the package with my own json schema profile and I was always getting this weird error "unhashable type: Row". This is because the jsonschema library tries to iterate over the elements of the Resource. 

# Solution
I think the best solution is to add an iterrows method that goes through the rows in the resource. And replace every place where one would iterate over the Resource with this method. 

I also considered converting the resource into a dictionary at "frictionless/metadata.py" line 274 but I think replacing the __iter__ method is not a good idea at all, please prove me wrong if you had a good reason to do this.

areleu:
- fixes #1227 

This is my proposal to solve an issue with validating Resource objects with a custom json schema.


roll:
Thanks!

I'm closing as per https://github.com/frictionlessdata/frictionless-py/issues/1227#issuecomment-1214883557

PR #1244, opened 2022-09-01
Added rule to filter numeric string with leading zero

### LINKED ISSUE #1232: Adding a field inferencing "subrule", otherwise very often I would have wrong extracted values
Hi,
daily I work with data where some columns have only numerical values in the cells, but which are actually string fields.

A typical case is that of administrative codes, of regions, provinces and cities (in the table below `codice_regione`, `codice_provincia` and `codice_comune` fields). In Italy, the city of Losine has the code `0870`, and obviously it's a string, and it must remain a string, otherwise I would lose the chance to JOIN this data with other data.

| comune | codice_regione | codice_provincia | codice_comune | denominazione_comune | sigla_provincia | data_entrata_in_carica |
| --- | --- | --- | --- | --- | --- | --- |
| 030151360 | 03 | 015 | 1360 | POLPENAZZE DEL GARDA | BS | 13/10/2021 |
| 030120530 | 03 | 012 | 0530 | CAROBBIO DEGLI ANGELI | BG | 04/10/2021 |
| 020040580 | 02 | 004 | 0580 | SAINT-DENIS | AO | 23/09/2020 |
| 030150870 | 03 | 015 | 0870 | LOSINE | BS | 04/10/2021 |
| 190480090 | 19 | 048 | 0090 | CAPO D'ORLANDO | ME | 27/10/2021 |

But if I run `extract` on it

```
frictionless extract "https://gist.githubusercontent.com/aborruso/076e5fad847b658a535b16cbcf3abdfd/raw/887edc88ab75b01d15ea1ac4bb052ffaf8d5ef9c/tmp.csv" --csv
```

the `codice_regione`, `codice_provincia` and `codice_comune` fields are no longer strings, they are all numbers and the cell values have changed (in example the code of Losine becomes `870`).

| comune | codice_regione | codice_provincia | codice_comune | denominazione_comune | sigla_provincia | data_entrata_in_carica |
| --- | --- | --- | --- | --- | --- | --- |
| 30151360 | 3 | 15 | 1360 | POLPENAZZE DEL GARDA | BS | 13/10/2021 |
| 30120530 | 3 | 12 | 530 | CAROBBIO DEGLI ANGELI | BG | 04/10/2021 |
| 20040580 | 2 | 4 | 580 | SAINT-DENIS | AO | 23/09/2020 |
| 30150870 | 3 | 15 | 870 | LOSINE | BS | 04/10/2021 |
| 190480090 | 19 | 48 | 90 | CAPO D'ORLANDO | ME | 27/10/2021 |

I know, octal numbers exist, but in 99% of cases (I'm talking about my experience), if I have fields with cells starting with zero they are not octal numbers, but string codes.

If it were possible I would add a subrule for all fields mapped as numbers: if there are cells that start with a zero,  not followed by a `,` or `.`, set that inferenced field as `string` and not as `number`.

Thank you.

shashigharti:
Added rule to prevent numeric string to be parsed as int
Added new test
Fixed previous tests

- fixes #1232

---


frictionless/fields/integer.py
```diff
@@ -36,6 +36,10 @@ def value_reader(cell):
             if isinstance(cell, str):
                 if pattern:
                     cell = pattern.sub("", cell)
+                # for numeric string starting with zero for example, 001, 00, 01
+                has_leading_zero = re.match(r"0\d+", cell.strip())
```
roll: Let's not use regural expressions here and in the number as it might hurt performance. 
This line will be executed 1 000 000 times for a numeric table with e.g. 100 x 10 000 dimension

frictionless/fields/year.py
```diff
@@ -25,6 +26,10 @@ def value_reader(cell):
                     return None
                 if len(cell) != 4:
                     return None
+                # for numeric string starting with zero for example, 001, 00, 01
+                has_leading_zero = re.match(r"0\d+", cell.strip())
```
roll: I think we should allow leading zeroes for `year` as the standards say that it should be in `YYYY` format - https://specs.frictionlessdata.io/table-schema/#year

So I think we just need to apply it for integer/number

roll (changes requested):
It looks good! 

I added a few comments

frictionless/fields/year.py
shashigharti: But for the example shared in the issue has a field "codice_comune" which I think is also a code but has the pattern of YYYY. The field gets inferred as year. But when I extract data, the zeros gets stripped:

- name: codice_comune
      type: year

`frictionless extract "https://gist.githubusercontent.com/aborruso/076e5fad847b658a535b16cbcf3abdfd/raw/887edc88ab75b01d15ea1ac4bb052ffaf8d5ef9c/tmp.csv"`

`codice_comune 
1360   
530
580
870
90

What do we do for such kind of field?

shashigharti:
@roll Thank you!. I have made the new changes and it is ready for the review.

frictionless/fields/integer.py
```diff
@@ -36,6 +36,13 @@ def value_reader(cell):
             if isinstance(cell, str):
                 if pattern:
                     cell = pattern.sub("", cell)
+                # for numeric string starting with zero for example, 001, 00, 01
+                cell_s = cell.strip()
+                has_leading_zero = (
+                    cell_s.startswith("0") and len(cell_s) > 1 and cell_s.isnumeric()
```
roll: I think we don't need `isnumeric` check here (if it's not not numeric it will fail later)

frictionless/fields/integer.py
```diff
@@ -36,6 +36,13 @@ def value_reader(cell):
...
+                )
+                if self.bare_number and has_leading_zero:
```
roll: We can put all the leading zeroes logic under the `if self.bare_number` for performance reasons

frictionless/fields/number.py
```diff
@@ -64,6 +64,13 @@ def value_reader(cell):
                 Primary = float
                 Secondary = Decimal
             if isinstance(cell, str):
+                # for numeric string starting with zero for example, 001, 00, 01
+                cell_s = cell.strip()
+                has_leading_zero = (
+                    cell_s.startswith("0") and len(cell_s) > 1 and cell_s.isnumeric()
```
roll: I think we don't need isnumeric check here (if it's not not numeric it will fail later)

frictionless/fields/number.py
```diff
@@ -64,6 +64,13 @@ def value_reader(cell):
...
+                )
+                if self.bare_number and has_leading_zero:
```
roll: We can put all the leading zeroes logic under the `if self.bare_number` for performance reasons

roll (changes requested):
Thanks @shashigharti just a few small comments

roll (approved):
Thanks!

shashigharti:
@roll thank you!. It is ready for the review.

PR #1247, opened 2022-09-02
Fix field_update with formula argument

l-vincent-l:
Before this fix, the content of all `population` cells would have been `None`.

I have to make a switch on `pass_row` to make it work with function.


shashigharti:
Thanks! @l-vincent-l . 
@roll  This looks good to me!

roll (approved):
Thanks @l-vincent-l @shashigharti !

PR #1259, opened 2022-09-26
Fixed casting int to string

### LINKED ISSUE #1251: Cast int cell to string in excel file
Let's say I want to parse the attached file.
It has in the A column `string` or `int`, but I want every cells of that column to be `string`.

I tried with this package.yaml

```yaml
resources:
  - name: example
    type: table
    path: example.xlsx
    scheme: file
    format: xlsx
    encoding: utf-8
    mediatype: application/vnd.ms-excel
    schema:
      fields:
        - name: A
          type: string
        - name: B
          type: string
    pipeline:
      steps:
        - type: table-normalize
        - type: field-merge
          fromNames:
            - A
            - B
          name: AB
```

and runned this python script

```python
from frictionless import Package

p = Package.from_descriptor('package-issue.yaml')

p.extract()

for r in p.resources:
    r.transform()
    print(r.to_view())
```

I have this error when I try to tranform it with `field-merge`
```
frictionless.exception.FrictionlessException: [step-error] Step is not valid: "field_merge" raises "sequence item 0: expected str instance, NoneType found" 
```

Is there a way to cast all ints in `string` ?

[example.xlsx](https://github.com/frictionlessdata/framework/files/9526532/example.xlsx)

shashigharti:
* Converts cell data to string while reading as done by csv reader

- fixes #1251 

---


shashigharti:
@roll it is ready for the review

roll (changes requested):
Hi @shashigharti,

I think it definitely needs to be a part of `ExcelControl` allowing a user to decide whether excel cells should be typed or not. Something like this:

```
resource = Resource('data.xlsx', control=controls.ExcelControl(stringified=True))
```

WDYT?

shashigharti:
> Hi @shashigharti,
> 
> I think it definitely needs to be a part of `ExcelControl` allowing a user to decide whether excel cells should be typed or not. Something like this:
> 
> ```
> resource = Resource('data.xlsx', control=controls.ExcelControl(stringified=True))
> ```
> 
> WDYT?

@roll Thank you! Yes I agree and have made changes accordingly. It is ready for review.

roll (approved):
Thanks!

roll:
@shashigharti 
I've added a few follow-ups:
- https://github.com/frictionlessdata/framework/commit/4f8217f5dce2e3d97df37464af0778c74502f506 (just to optimize a little bit)
- https://github.com/frictionlessdata/framework/commit/8adb942c4a0cc8231904fbf1a84ccc5bc1fee1c0

PR #1264, opened 2022-10-14
Fixed dialect for extract

### LINKED ISSUE #1009: Extract CSV, with field delimiter consistent with the schema used
Please replace this line with full information about your idea or problem. If it's a bug share as much as possible to reproduce it.

I have this input CSV file:

```
fieldNameA;fieldNameB
0123;c
```

If I run `extract` the field delimiter is changed from `;` to `,`:

```
fieldNameA,fieldNameB
123,c
```

I think that the default way should be to have the output field separator as the one of the input.

Moreover if I have this schema, in which I set `delimiter: ;`

```yaml
path: input.csv
name: input
scheme: file
format: csv
hashing: md5
encoding: utf-8
dialect:
  delimiter: ;
schema:
  fields:
    - type: integer
      name: fieldNameA
    - type: string
      name: fieldNameB
```

and then run `frictionless extract input.yml --csv`, I have again `,` as field separator and it's inconsistent with the schema I have.

```
fieldNameA,fieldNameB
123,c
```

I think that if I use a schema, I should have a consistent output also for delimiter setting.
This is related to this [discussion](https://github.com/frictionlessdata/project/discussions/715#discussioncomment-2412867=).

Thank you

shashigharti:
- Added bug fix for CSV delimiter. Now it does not change the delimiter to default ","
- fixes #1009

---


shashigharti:
@roll this is ready for review

tests/program/test_extract.py
```diff
@@ -293,3 +293,48 @@ def test_program_extract_single_resource_csv():
...
+    assert (
+        json.dumps(actual.stdout, ensure_ascii=False)
+        == '"fieldNameA;fieldNameB\\n123;c\\n"'
+    )
+
+
+def test_extract_resource_from_csv_comma_delimiter_param_issue_1009():
+    actual = runner.invoke(
+        program,
+        'extract data/issue-1009-comma.csv --dialect \'{"delimiter": ","}\' --csv',
```
roll: Unfortunately, we can't do something like this. `extract --dialect` means "specify a dialect of the input file" but not for the output one.

roll (changes requested):
Thanks, @shashigharti,

It's a really great iteration to better understand the problem but I'm afraid we need to think twice about this. The problem here is that we're stepping into a really big area of data conversion when we have input options and output options. 

I know that it was in the initial feature request (and it's not a bug btw) but I started thinking more and more about this and it feels that first we will keep the delimiter than using the same logic we need to keep the header rows (e.g. starting from row 2) etc. So basically, it will break the idea of the `extract` actions as a normalization function.

I would suggest reconsidering the implementation and implement it as a new feature that doesn't affect current behavirou. For example, what if we keep everything existent as it is BUT add a new flag like `keep-delimiter` for `program.extract` that will implement the feature request?

WDYT?

shashigharti:
> Thank you! I agree with that. I will try to change the implementation considering your suggestions.

shashigharti:
@roll thank you!, I have revised the code.  And now it will retain the delimiter if dialect info has been specified in params and --keep-params is set to true.

It is ready for the review.

roll (approved):
Amazing! Thanks!

aborruso:
thank you very much @shashigharti and @roll

PR #1265, opened 2022-10-14
Fixed bug in table-write step while using foreign key

### LINKED ISSUE #1252: Unable to do a table-write of a table with foreignkeys to a postgres database
If I have this `package.yaml` file
```yaml
name: petites_villes_de_demain
resources:
  - name: commune
    dialect:
      sql:
        table: commune
    format: sql
    mediatype: application/sql
    path: postgresql://user:pwd@localhost/mydb
    type: table
    schema:
      fields:
      - name: insee_com
        type: string
        constraints:
          required: true
      primaryKey:
      - insee_com
  - name: diffusion-zonages-zrr-cog2021
    type: table
    path: https://www.data.gouv.fr/fr/datasets/r/4160134c-017c-42c7-b838-6048ad56e5f2
    scheme: https
    format: xls
    encoding: utf-8
    mediatype: application/vnd.ms-excel
    dialect:
      headerRows:
        - 6
    schema:
      fields:
        - name: CODGEO
          type: string
        - name: LIBGEO
          type: string
        - name: ZRR_SIMP
          type: string
        - name: ZONAGE_ZRR
          type: string
      foreignKeys:
        - fields: CODGEO
          reference:
            resource: commune
            fields:
              - insee_com

    pipeline:
      steps:
      - type: table-write
        path: postgresql://user:pwd@localhost/mydb
        dialect:
              sql:
                table: zrr
```

*TLDR* This file describes a package with 2 resources : 
 * one is the list of towns if France, it's stored in a postgresql table
 * the other one describes data linked to towns, and it has a pipeline that writes this table to the postgresql datababase.
 
 Currently this doesn't work because when we called the `table-write` on the resource we don't have knowledge of the original package anymore.
 To do the transformation there is a creation of a new package [here](https://github.com/frictionlessdata/framework/blob/main/frictionless/formats/sql/storage.py#L181), so when we prepare the lookups we don't find the resource we are looking for, and it raises an exception [here](https://github.com/frictionlessdata/framework/blob/main/frictionless/resource/resource.py#L720)

shashigharti:
- It fixes the bug while writing a table/resource using resource pipeline 'steps' 'table-write'. The bug appeared when the table being written had foreign key reference with another table resource.
-  There were bugs in two steps:
1. Please provide "dialect.sql.table" for writing
2. Error in __prepare_lookup() [here](https://github.com/frictionlessdata/framework/blob/5e092abae1167c6d8e452cb934b9f202ecf28801/frictionless/resource/resource.py#L720)
- fixes #1252

Need feedback as I am not sure if the solution that I have given is the right way.

---

### DIFF
--- a/data/package-with-pipeline-multiple-tables-to-write.json
+++ b/data/package-with-pipeline-multiple-tables-to-write.json
@@ -1,0 +1,151 @@
+{
+  "name": "petites_villes_de_demain",
+  "resources": [
+    {
+      "name": "commune",
+      "type": "table",
+      "path": "sqlite:///database.db",
+      "format": "sql",
+      "mediatype": "application/sql",
+      "dialect": {
+        "sql": {
+          "table": "commune"
+        }
+      },
+      "schema": {
+        "fields": [
+          {
+            "name": "insee_com",
+            "type": "string",
+            "constraints": {
+              "required": true
+            }
+          }
+        ],
+        "primaryKey": [
+          "insee_com"
+        ]
+      }
+    },
+    {
+      "name": "diffusion-zonages-zrr-cog2021",
+      "type": "table",
+      "path": "https://github.com/fdtester/temp-files/blob/main/diffusion-zonages-zrr-cog2021.xls?raw=true",
+      "scheme": "https",
+      "format": "xls",
+      "encoding": "utf-8",
+      "mediatype": "application/vnd.ms-excel",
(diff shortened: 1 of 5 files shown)


shashigharti:
@roll looking for your feedback as I am not very sure of the method that I have used to solve it.

roll (approved):
Thanks!

PR #1275, opened 2022-10-26
1261/fixes serialization of tabular data profile

### LINKED ISSUE #1261: Cannot serialize package descriptor with tabular data profile
We’re using Frictionless framework v5 beta.

We’d like to serialize a package descriptor with an explicit `tabular-data-package` profile.

Appartently, it’s not possible to serialize a resource descriptor with a `data-resource` or `tabular-data-resource` profile, or a package descriptor with the `data-package` or `tabular-data-package` profile.

The serialization fails with: `frictionless.exception.FrictionlessException: [metadata-error] Metadata error: cannot retrieve metadata "tabular-data-resource" because "[Errno 2] No such file or directory: 'tabular-data-resource'"`

The issue happens in `metadata_validate`, where the code expects the profile to be the path to a schema.

Should builtin profiles like  `data-resource` or `tabular-data-resource` be treated in a specific way?

Here are some tests to demonstrate the issue:
```python
@pytest.mark.parametrize("profile", ["data-resource", "tabular-data-resource"])
def test_data_resource_profile(profile):
    resource = Resource(
        name="table", path="data/table.csv", profiles=[profile]
    )
    descriptor = resource.to_descriptor()
    assert descriptor["profile"] == profile


@pytest.mark.parametrize("profile", ["data-package", "tabular-data-package"])
def test_data_package_profile(profile):
    resource = Resource(name="table", path="data/table.csv")
    package = Package(resources=[resource], profiles=[profile])
    descriptor = package.to_descriptor()
    assert descriptor["profile"] == profile
```

shashigharti:
- When we set profiles as ["tabular-data-resource", "data-resource", "Other"], it is not handled in the code (line 284, metadata.py). It reads it as file path and tries to open the file and error occurs. To fix that, this fix adds the condition, which checks for the profile if it is path, remotefile or a dict.
- fixes #1261

---


shashigharti:
@roll this is ready for the review

roll (changes requested):
Hi @shashigharti,

We need to handle compatibility between v4 and v5 in:
- IMPORT: `package.metadata_transform` function (see `resource.metadata_transform`)
- EXPORT: `package.metadata_export` (not related to this issue)

But for the Package the problem is bigger. Profiles like `data-package` and `tabular-data-package` are basically useless with a new architecture. I think the only thing we can do is:
- to ignore/remove `data-package` profile in `package.metadata_transform`
- set all the resource to `type=table` if a profile is `tabular-data-package`

shashigharti:
> * to ignore/remove `data-package` profile in `package.metadata_transform`
> * set all the resource to `type=table` if a profile is `tabu`

@roll  Thank you!. I have made the changes accordingly and updated the PR. It is ready for review.

PR #1338, opened 2022-11-29
Fix for timestamp comparison

### LINKED ISSUE #1210: Pandas warning "Comparison of Timestamp"
# Overview

In v5 branch for this test:

```
def test_pandas_parser_write_types():
    source = Package("data/storage/types.json").get_resource("types")
    target = source.write(format="pandas")
    with target:

        # Assert schema
        assert target.schema.to_descriptor() == {
            "fields": [
                {"name": "any", "type": "string"},  # type fallback
                {"name": "array", "type": "array"},
                {"name": "boolean", "type": "boolean"},
                {"name": "date", "type": "datetime"},
                {"name": "date_year", "type": "datetime"},  # format removal
                {"name": "datetime", "type": "datetime"},
                {"name": "duration", "type": "duration"},
                {"name": "geojson", "type": "object"},
                {"name": "geopoint", "type": "array"},
                {"name": "integer", "type": "integer"},
                {"name": "number", "type": "number"},
                {"name": "object", "type": "object"},
                {"name": "string", "type": "string"},
                {"name": "time", "type": "time"},
                {"name": "year", "type": "integer"},  # type downgrade
                {"name": "yearmonth", "type": "array"},  # type downgrade
            ],
        }

        # Assert rows
        assert target.read_rows() == [
            {
                "any": "中国人",
                "array": ["Mike", "John"],
                "boolean": True,
                "date": date(2015, 1, 1),
                "date_year": date(2015, 1, 1),
                "datetime": datetime(2015, 1, 1, 3, 0),
                "duration": isodate.parse_duration("P1Y1M"),
                "geojson": {"type": "Point", "coordinates": [33, 33.33]},
                "geopoint": [30, 70],
                "integer": 1,
                "number": 7,
                "object": {"chars": 560},
                "string": "english",
                "time": time(3, 0),
                "year": 2015,
                "yearmonth": [2015, 1],
            },
        ]
```

We get a warning:

```
 FutureWarning: Comparison of Timestamp with datetime.date is deprecated in order to match the standard library behavior. In a future version these will be considered non-comparable. Use 'ts == pd.Timestamp(date)' or 'ts.date() == date' instead.
    return super().__eq__(other)
```

shashigharti:
- resource->write function did not pass schema to new resource created and the fields inferred would have different type (due to inferred field type). 
- So if the schema is specified, it will now be passed. 
- Also changed the type of geopoint and yearmonth to string  in types.json file.
- fixes #1210

---

### DIFF
--- a/data/storage/types.json
+++ b/data/storage/types.json
@@ -12,14 +12,14 @@
           {"name": "datetime", "type": "datetime"},
           {"name": "duration", "type": "duration"},
           {"name": "geojson", "type": "geojson"},
-          {"name": "geopoint", "type": "geopoint"},
+          {"name": "geopoint", "type": "string"},
           {"name": "integer", "type": "integer"},
           {"name": "number", "type": "number"},
           {"name": "object", "type": "object"},
           {"name": "string", "type": "string"},
           {"name": "time", "type": "time"},
           {"name": "year", "type": "year"},
-          {"name": "yearmonth", "type": "yearmonth"}
+          {"name": "yearmonth", "type": "string"}
         ]
       },
       "data": [
--- a/frictionless/resource/resource.py
+++ b/frictionless/resource/resource.py
@@ -952,6 +952,8 @@ class Resource(Metadata):
         """
         resource = target
         if not isinstance(resource, Resource):
+            if self.has_schema:
+                options["schema"] = self.schema
             resource = Resource(target, control=control, **options)
         resource.infer(sample=False)
         parser = system.create_parser(resource)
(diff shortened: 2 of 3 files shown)


shashigharti:
@roll this is ready for review. I am not very sure if the solution is right, looking for your feedback. Thanks!

roll:
@shashigharti 
I'm closing in favor of #1374. I investigated and I think that the only problem was that we didn't convert `pd.Timestamp` to `datetime` on reading from Pandas

PR #1363, opened 2023-01-04
support for string format "wkt"

jze:
I have added support for a new string format "wkt". [Well-Known-Text](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry) (WKT) is a common way to write down geographic data. Now you can specify a column containing WKT like this:

```json
{
  "name": "geometry",
  "type": "string",
  "format": "wkt"
}
```

The WKT content will be validated. The validation uses the WKT grammar that can be found here: https://git.osgeo.org/gitea/postgis/postgis/src/branch/master/doc/bnf-wkt.txt I have converted it into a syntax that be used with [grako](https://pypi.org/project/grako/). I have put that file to `data/wkt.txt` because I didn't know a better place to put it. The `frictionless/wkt_parser.py` has been generated from this file:

```bash
grako wkt.txt --name Wkt -o frictionless/wkt_parser.py
```

The support for the WKT format is mentioned in this issue: https://github.com/frictionlessdata/specs/issues/772


roll (approved):
Amazing @jze !

Thanks! Also, do you think WKT is only helpful as a string format or it can have its own type?

jze:
Thank you for merging!

Somehow an own type make sence since geometries are somehow 'special'. However, since the types are based on the type set of json-schema it might be better to stick with the small selection of types as it is now.

roll:
Thanks!

I think it's anyway a good start and for the own type conversation I mean that for the framework the difference is made by how values are represented in Python. So, currently, it's definitely a string type as they are string in Python. But if there were some Python library to represent WKT or it was possible to have it as a dict/json it might be another type

PR #1428, opened 2023-02-17
Add file_create endpoint

shashigharti:
- fixes https://github.com/frictionlessdata/application/issues/111

---


frictionless/project/filesystem.py
```diff
@@ -48,7 +48,7 @@ def copy_file(self, path: str, *, folder: Optional[str] = None) -> str:
         return path
 
     # TODO: use streaming?
-    def create_file(
+    def upload_file(
```
roll: @shashigharti 
Actually, we don't need to do so we can keep just one `filesystem.create_file` endpoint as it already accepts bytes and use it in both:
- `project.create_file`
- `project.upload_file`

frictionless/project/filesystem.py
```diff
@@ -60,6 +60,17 @@ def create_file(
...
         return path
 
+    def create_file(self, path: str, *, folder: Optional[str] = None) -> str:
+        if folder:
+            folder = self.get_secure_fullpath(folder)
+            assert self.is_folder(folder)
+        resource = Resource(path)
+        name = str(path.split("/")[-1])
+        path = self.get_secure_fullpath(folder, name, deduplicate=True)
+        resource.write(path)
```
roll: Instead of `resource.write` it will be better to `filesystem.create_file(bytes=resource.read_bytes())` (see above as well)

frictionless/server/file/create.py
```diff
@@ -17,12 +17,10 @@ class Result(BaseModel):
 @router.post("/file/create")
 async def server_file_create(
```
roll: @shashigharti 
OK, the problem is here -- `/file/create` endpoint should be just a normal JSON endpoint (see the way others endpoints are written with `Props/Result` data structures)

The reason why `/file/upload` is written differently to other endpoint is because it needs to handle file uploading - https://fastapi.tiangolo.com/tutorial/request-files/

Also note, that for file uploading client's `makeRequest` does some magic with multipart form

shashigharti: Thank you!

shashigharti:
Thank you @roll!. It is ready for the review.

roll (approved):
Thanks!

PR #1454, opened 2023-03-06
Add export and download endpoint for table

shashigharti:
- fixes #[81](https://github.com/frictionlessdata/application/issues/81)

---


shashigharti:
@roll, Can you please provide your feedback on it.  Thank you!

I tried using [here](https://github.com/frictionlessdata/framework/pull/1454/commits/4b8cb050c0bd1e386e0b9aa92078d75f49f1ddce#diff-d6eda36f1fd700c4e92b9f88142261709f0cf8c924753de9470ae9e4a7126295R215)
`Resource(sql_database, control={sql_table}).write(filename.format)`

but it adds metadata _rowNumber which complains while indexing (about primary key)

frictionless/project/database.py
```diff
@@ -253,6 +253,10 @@ def update_record(self, path: str):
 
     # Table
 
+    def export_table(self, path: str, newfilepath: str) -> Resource:
+        resource = Resource(path=path).write(newfilepath)
```
roll: Can you please just add a TODO that it's a placeholder code? We need to export from the database but obviously I need to fix the underlying logic to make it work..

frictionless/server/table/export.py
```diff
@@ -0,0 +1,23 @@
+from typing import Optional
+from pydantic import BaseModel
+from fastapi import Request
+from ...project import Project
+from ..router import router
+
+
+class Props(BaseModel):
+    session: Optional[str]
+    path: str
```
roll: It seems we can just accept `source` (path) and `target` (path) as we only use format to merge it with the name at the beginning (and the client can merge them by itself)

roll (commented):
Thanks a think it's a good starter I added a few comments

shashigharti:
@roll Thank you for the feedback!. I have made changes according to your suggestion.

roll (approved):
Thanks!

PR #1469, opened 2023-03-27
Write package file in ckan publish

### LINKED ISSUE #1463: CKAN publishing doesn't seem to be writing `datapackage.json`
# Overview

It's been published with Frictionless and there is no `datapackage.json` (and the whole idea is that we publish data + metadata)

ckan

shashigharti:
- Add test
- Add package file upload
- Fix tests
- fixes #1463

---


roll:
Thanks @shashigharti!

Can you please test it for a dataset like this (resources in a folder -- we might need to rewrite the paths):

```
$frictionless: package/v2
resources:
  - name: chunk1
    type: table
    path: data/tables/chunk1.csv
    scheme: file
    format: csv
    mediatype: text/csv
    encoding: utf-8
    schema:
      fields:
        - name: id
          type: integer
        - name: name
          type: string
  - name: chunk2
    type: table
    path: data/tables/chunk2.csv
    scheme: file
    format: csv
    mediatype: text/csv
    encoding: utf-8
    schema:
      fields:
        - name: id
          type: integer
        - name: name
          type: string
```

shashigharti:
> Thanks @shashigharti!
> 
> Can you please test it for a dataset like this (resources in a folder -- we might need to rewrite the paths):
> 
> ```
> $frictionless: package/v2
> resources:
>   - name: chunk1
>     type: table
>     path: data/tables/chunk1.csv
>     scheme: file
>     format: csv
>     mediatype: text/csv
>     encoding: utf-8
>     schema:
>       fields:
>         - name: id
>           type: integer
>         - name: name
>           type: string
>   - name: chunk2
>     type: table
>     path: data/tables/chunk2.csv
>     scheme: file
>     format: csv
>     mediatype: text/csv
>     encoding: utf-8
>     schema:
>       fields:
>         - name: id
>           type: integer
>         - name: name
>           type: string
> ```

Thank you! @roll. I tested with above file and it was complaining about` 'data/data/tables/chunk2.csv' `not found. The error was while reading bytes: `resource.read_bytes()`. So I used this instead(path relative to package file).

```
$frictionless: package/v2
name: test-package-file-write-with-resources-inside-folder
title: Test Package File Write With Resource inside folder
resources:
  - name: chunk1
    type: table
    path: tables/chunk1.csv
    scheme: file
    format: csv
    mediatype: text/csv
    encoding: utf-8
    schema:
      fields:
        - name: id
          type: integer
        - name: name
          type: string
  - name: chunk2
    type: table
    path: tables/chunk2.csv
    scheme: file
    format: csv
    mediatype: text/csv
    encoding: utf-8
    schema:
      fields:
        - name: id
          type: integer
        - name: name
          type: string
```

I have added the test for the above (writing and reading full cycle). Looking for your feedback.
Thanks!

shashigharti:
@roll sorry I did not understand before your hint. I have made the changes to rewrite the file path in resources as per your feedback. Thank you!

frictionless/portals/ckan/adapter.py
```diff
@@ -106,9 +107,33 @@ def write_package(self, package: Package) -> Union[None, str]:
 
                 # upload resources
                 # TODO: See if it's possible to upload only the resources that need to be uploaded
-                for resource in package.resources:
+                for index, resource in enumerate(package.resources):
+                    if resource.path:
+                        resource_filename = resource.path.split("/")[-1]
+                        package.resources[index].path = resource_filename
```
roll: I think it might be better to get `descriptor = package.to_descriptor()` first and then do modifications as we don't want to affect initial package in general

frictionless/portals/ckan/adapter.py
```diff
@@ -106,9 +107,33 @@ def write_package(self, package: Package) -> Union[None, str]:
 
                 # upload resources
                 # TODO: See if it's possible to upload only the resources that need to be uploaded
-                for resource in package.resources:
+                for index, resource in enumerate(package.resources):
+                    if resource.path:
+                        resource_filename = resource.path.split("/")[-1]
```
roll: Probably it's better to use `os.path` or `pathlib.Path` to handle separators as the current code might fail on windows

frictionless/portals/ckan/adapter.py
roll: Also, it's better to use `package.to_descriptor(validate=True)` (it's a new flag) to ensure that the package is valid. For example, we might get into a problem of duplicate resource names:
- table.csv
- data/table.csv

Maybe it will be better to replace names like:
- table.csv -> table.csv
- data/table.csv -> data_table.csv

shashigharti: The name collision seems to be handled on CKAN part. It uses `name` of the resource to write the file. So I changed the path rewrite to use `name` .

shashigharti:
Thanks! @roll.I have made the changes as per your feedback.

roll (approved):
Thanks!

PR #1509, opened 2023-05-16
Update validating-data.md

sapetti9:
Links to Check Guide and API Reference are broken. @roll can you please check I updated the hyperlinks with the right links? Thanks.


roll (approved):
Thanks!

PR #1523, opened 2023-06-12
Fix to set header to false

### LINKED ISSUE #1513: Cannot create CSV/TSV file without header
# Problem Description

Hi there,
I was trying to create a CSV/TSV file WITHOUT header, but was not able to do it.

# Versions

Frictionless:
```
$ pip freeze | grep frictionless
frictionless==5.13.1
```
OS:
```
Microsoft Windows [Version 10.0.19045.2846]
```

# Sample code:
```
from frictionless import Dialect, Resource


def main():
    with Resource("capital-3.csv") as resource:
        print(f"Header: {resource.header}")
        print(f"Valid: {resource.header.valid}")

        new_table = "tmp_table.tsv"
        dialect = Dialect.from_descriptor({"header": False, "delimiter": "\t"})
        resource.write_table(new_table, dialect=dialect, encoding='utf-8')

    print()
    print("Result (expected WITHOUT header):")
    print(open(new_table).read())


if __name__ == "__main__":
    main()
```

I'm reading a simple CSV file with a header and two columns and writing it to another file. 

Note the 'header = False' in the dialect:

```dialect = Dialect.from_descriptor({"header": False, "delimiter": "\t"})```

# stdout
```
Header: ['id', 'name']
Valid: True

Result (expected WITHOUT header):
id	name
1	London
2	Prague
3	Berlin
```

# Current Result - 'tmp_table.tsv' file:
Header is present:
```
id	name
1	London
2	Prague
3	Berlin
```

# Expected Result - 'tmp_table.tsv' file:
Shall be WITHOUT header:
```
1	London
2	Prague
3	Berlin
```

# Conclusion
Is it a bug or a mistake of mine somewhere?

Regards,
Roman

### PR DESCRIPTION
**Pull request by shashigharti**
- fixes #1513

---

### CODE REVIEW

**Comment by shashigharti** [comment]
@roll it is ready for review. Thank you!. 

I have few concerns regarding the test that I have added. If `target` is already a table resource, then there is no way to set the `dialect` through write function. For example this does not work:

```
def test_yaml_parser_write_skip_header(tmpdir):
    dialect = Dialect.from_descriptor({"header": False})
    target = TableResource(path=str(tmpdir.join("table.yaml")))  # I did not set the dialect here
    with TableResource(path="data/table.csv") as resource:
        assert resource.header == ["id", "name"]
        resource.write_table(target, dialect=dialect) # This does not get applied
        assert target.read_rows() == [
            {"field1": 1, "field2": "english"},
            {"field1": 2, "field2": "中国人"},
        ]
```

File: tests/formats/excel/parsers/test_xls.py [hunk 1]
```diff
@@ -126,6 +127,18 @@ def test_xls_parser_write_sheet_name(tmpdir):
...
+def test_xls_parser_write_skip_header(tmpdir):
+    control = formats.ExcelControl(sheet="sheet")
+    dialect = Dialect.from_descriptor({"header": False})
+    data = b"header1,header2\nvalue11,value12\nvalue21,value22"
+    target = str(tmpdir.join("table.xls"))
+    with TableResource(data=data, format="csv") as resource:
+        assert resource.header == ["header1", "header2"]
+        resource.write_table(target, dialect=dialect, control=control)
+    with Resource(target) as resource:
+        assert resource.header == ["value11", "value12"]  # type: ignore
```
roll: If you use `TableResource` you don't need "type: ignore" here and in other tests

roll (changes requested):
Amazing work @shashigharti!

Just a small request change regarding typing

tests/formats/excel/parsers/test_xls.py
shashigharti: Thanks! @roll I tried that and I can omit `# type: ignore`  however I had faced another issue, the dialect does not get set using `write_table` [here](https://github.com/frictionlessdata/framework/blob/main/frictionless/resources/table.py#L454)
, so I had to use `string` path instead.
```
def test_xls_parser_write_skip_header(tmpdir):
    control = formats.ExcelControl(sheet="sheet")
    dialect = Dialect.from_descriptor({"header": False})
    data = b"header1,header2\nvalue11,value12\nvalue21,value22"
    target = TableResource(path=str(tmpdir.join("table.xls")), control=control)
    with TableResource(data=data, format="csv") as resource:
        assert resource.header == ["header1", "header2"]
        resource.write_table(target, dialect=dialect)
    with target:
        assert target.header == ["value11", "value12"]
```

roll: @shashigharti I mean just for reading:
```
with TableResource(data=data, format="csv") as resource:
        assert resource.header == ["header1", "header2"]
        resource.write_table(target, dialect=dialect, control=control)
    with TableResource(path=str(tmpdir.join("table.xls")), control=control) as resource:
        assert resource.header == ["value11", "value12"]
```

Or shall we fix `table.write_table` to support dialect?

shashigharti: Thanks @roll I have made the changes.

shashigharti:
Thanks! @roll, I have made the changes. It is ready for review

I tried this also to update other params along with dialect but I got `path is not safe` error:

`resource = resource.to_copy(control=control, **options)`

roll (approved):
Thanks!

PR #1531, opened 2023-06-19
Fix for missing values on excel in numeric format

### LINKED ISSUE #1503: Missing value on a date column on excel throws validation error
# Overview

I'm trying to validate a resource that stores data in an excel file in which of of the columns should be a date but missing values coded as 0 should be allowed. Here's the data:

20230512T092224

and the table schema

```json
      "schema": {
        "fields": [
          {
            "name": "id",
            "type": "integer"
          },
          {
            "name": "updated_at",
            "type": "date",
            "missingValues": [
              "0"
            ]
          }
        ]
      }
```

However, when I validate this resource I get a type error saying that

```
Type error in the cell "0" in row "3" and field "updated_at" at position "2": type is "date/default"
```

For a resource using a csv file to store data the same schema works.

There is a reproducible example available [here](https://github.com/splor-mg/datapackage-reprex/tree/main/20230512T084359).

shashigharti:
- fixes #1503

---


shashigharti:
@roll it is ready for the update. Thank you!

roll (approved):
Thanks!

PR #1541, opened 2023-06-23
Constraint one column on another

### LINKED ISSUE #1521: Constrain one column depending on another
# Overview

Hi, I've searched accross the other issues and I haven't been able to find a similar topic: within the framework of the [schema.data.gouv.fr](https://schema.data.gouv.fr/) initiative, we have been looking for a way to constrain one column depending on another one in a TableSchema. For example:
- `col1` is constrained as a boolean
- we would like `col2` to be constrained as follows: if `col1` is `true`, `col2` is restrained to `['1', '2', '3']` ; if `col1` is `false`, `col2` is restrained to `['A', 'B', 'C']`

A slightly more complex case would be that `col1` is constrained with `enum` and then for each different value that `col1` can have, `col2` has a different set of `enum` constraints.
Is there a way to do so in the current state of the TableSchema standard?

shashigharti:
- fixes #1521 
- The fix adds `row-constraint` in the checklist scope which was missing. We can now use the custom code to add row constraint as given in the added test
- Another solution would be to add additional fix to `row_constraint` step to include the function definitions in the `simpleeval`  

---

### DIFF
--- a/frictionless/checks/baseline.py
+++ b/frictionless/checks/baseline.py
@@ -42,6 +42,7 @@ class baseline(Check):
         errors.BlankRowError,
         errors.PrimaryKeyError,
         errors.ForeignKeyError,
+        errors.RowConstraintError,
         # Cell
         errors.ExtraCellError,
         errors.MissingCellError,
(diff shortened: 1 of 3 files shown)


shashigharti:
@roll it is ready for review. Thank you!

shashigharti:
Used another solution, closing PR

PR #1586, opened 2023-09-02
added libraries so that `make lint` inside the Docker container works

### LINKED ISSUE #1585: Running the tests inside the Docker container does not work
Running the tests for frictionless-py using Docker as described in <https://github.com/frictionlessdata/frictionless-py/blob/main/CONTRIBUTING.md#using-docker> does not work. There are apparently some libraries missing for Ubuntu 22.04. The  `hatch` command is also missing in the container.

## Steps to reproduce

```
make docker
docker run --rm -v $PWD:/home/frictionless -it frictionless-dev
make lint
```

jze:
- fixes #1585


roll (approved):
Thanks @jze !

PR #1587, opened 2023-09-03
Replaced the outdated grako library to parse WKT with TatSu

### LINKED ISSUE #1584: Replace grako with something else
The grako library is used to process WKT. However, it is not maintained anymore. It is already excluded if using Python 3.10. Therefore, it must be replaced with something else.

jze:
I have replaced the outdated grako library with the new [TatSu](https://tatsu.readthedocs.io) parser generator. TatSu also works with Python 3.10. Therefor, the tests for wkt format could be re-enabled.

- fixes #1584


roll (approved):
Thanks @jze !

PR #1588, opened 2023-09-03
additional tests with ods file containing more than one sheet

jze:
I have added a ods file that contains more than one sheet. The tests make sure that information from the 2nd sheet can be read as well.


roll (approved):
Thanks @jze !

PR #1594, opened 2023-10-01
Fix mypy error `error: invalid syntax [syntax]`

### LINKED ISSUE #1593: mypy: `error: invalid syntax  [syntax]` at `frictionless/dialect/types.py:28`
Dear @roll,

first things first: Thank you so much for conceiving and maintaining Frictionless. It is true gold.

At https://github.com/daq-tools/skeem/pull/10, we believe we are being tripped by a similar issue already reported at GH-1580 by @aborruso.

At our end, it is the mypy type checker, which fails like:
```
$ mypy skeem
.venv/lib/python3.10/site-packages/frictionless/dialect/types.py:28: error: invalid syntax  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```
... when upgrading to frictionless-5.14.x already:
```
pip install --upgrade "frictionless<5.15"
```

On the other hand, when downgrading to frictionless-5.13.x:
```
pip install --upgrade "frictionless<5.14"
```

... mypy succeeds again:
```
mypy --install-types --non-interactive skeem
Success: no issues found in 36 source files
```

With kind regards,
Andreas.

P.S.: We are using the most recent mypy 1.5.1:
```
$ mypy --version
mypy 1.5.1 (compiled: yes)
```

amotl:
Hi again,

this patch intends to fix #1593. After applying it manually, the mypy type checker was able to work past this error. Apparently, it fails on the inline comments?

With kind regards,
Andreas.

### DIFF
--- a/frictionless/resource/types.py
+++ b/frictionless/resource/types.py
@@ -11,7 +11,6 @@ from ..schema import ISchema
 # TODO: replace by frictionless-standards
 class IResource(TypedDict, total=False):
     name: Required[str]
-    #  type = required
     title: str
     description: str
     path: str
--- a/frictionless/schema/types.py
+++ b/frictionless/schema/types.py
@@ -18,7 +18,6 @@ class ISchema(TypedDict, total=False):
 
 class IField(TypedDict, total=False):
     name: Required[str]
-    #  type = required
     title: str
     description: str
     format: str


roll (approved):
Thanks a lot!

PR #1620, opened 2024-01-15
Fix 1611 incorrect errors when validating data with a missiing required column

### LINKED ISSUE #1611: Incorrect errors when validating data with a missing required column
# Overview

In the process of migration from v4 to v5 in validata, we experienced some incorrect errors in the case of a missing required column. 

Here is some python code to reproduce : 

```python
import frictionless

source = [["B", "C"], ["b", "c"]]
schema = {
    "$schema": "https://frictionlessdata.io/schemas/table-schema.json",
    "fields": [
        {
            "name": "A",
            "title": "Field A",
            "type": "string",
            "constraints": {"required": True},
        },
        {"name": "B", "title": "Field B", "type": "string"},
        {"name": "C", "title": "Field C", "type": "string"},
    ],
}

if __name__ == "__main__":
    schema = frictionless.Schema.from_descriptor(schema)
    resource = frictionless.Resource(
        source, schema=schema, detector=frictionless.Detector(schema_sync=True)
    )
    report = frictionless.validate(resource)
    print(report)
```

Output : 

```
{'valid': False,
 'stats': {'tasks': 1, 'errors': 3, 'warnings': 0, 'seconds': 0.005},
 'warnings': [],
 'errors': [],
 'tasks': [{'name': 'memory',
            'type': 'table',
            'valid': False,
            'place': '<memory>',
            'labels': ['B', 'C'],
            'stats': {'errors': 3,
                      'warnings': 0,
                      'seconds': 0.005,
                      'fields': 3,
                      'rows': 1},
            'warnings': [],
            'errors': [{'type': 'missing-label',
                        'title': 'Missing Label',
                        'description': 'Based on the schema there should be a '
                                       "label that is missing in the data's "
                                       'header.',
                        'message': "There is a missing label in the header's "
                                   'field "A" at position "3"',
                        'tags': ['#table', '#header', '#label'],
                        'note': '',
                        'labels': ['B', 'C'],
                        'rowNumbers': [1],
                        'label': '',
                        'fieldName': 'A',
                        'fieldNumber': 3},
                       {'type': 'constraint-error',
                        'title': 'Constraint Error',
                        'description': 'A field value does not conform to a '
                                       'constraint.',
                        'message': 'The cell "None" in row at position "2" and '
                                   'field "A" at position "3" does not conform '
                                   'to a constraint: constraint "required" is '
                                   '"True"',
                        'tags': ['#table', '#row', '#cell'],
                        'note': 'constraint "required" is "True"',
                        'cells': ['b', 'c'],
                        'rowNumber': 2,
                        'cell': 'None',
                        'fieldName': 'A',
                        'fieldNumber': 3},
                       {'type': 'missing-cell',
                        'title': 'Missing Cell',
                        'description': 'This row has less values compared to '
                                       'the header row (the first row in the '
                                       'data source). A key concept is that '
                                       'all the rows in tabular data must have '
                                       'the same number of columns.',
                        'message': 'Row at position "2" has a missing cell in '
                                   'field "A" at position "3"',
                        'tags': ['#table', '#row', '#cell'],
                        'note': '',
                        'cells': ['b', 'c'],
                        'rowNumber': 2,
                        'cell': '',
                        'fieldName': 'A',
                        'fieldNumber': 3}]}]}
```

## Observed behavior

There are three errors among which :
- there is a `Constraint Error` that suggests that a field  `A` has a missing value (`None`) although there is no column A at all. 
- there is a `Missing Cell` error that suggests that input data is malformed, although input data is perfectly fine. 

## Expected behavior

I would expect to only get the first `missing-label` error. 

## Other details and experimentations

Frictionless version 5.16.0

Same result with command line validation.
I have put "schema-sync" to reproduce more closely our use case, but it does not seem to be related with the actual issue. 

Inspecting "row" on ["validator/validator.py", l151](https://github.com/frictionlessdata/frictionless-py/blob/main/frictionless/validator/validator.py#L151) :
```python
                        row = next(resource.row_stream)  # type: ignore
```
returns an artificially added A property :
```python
{'B': 'b', 'C': 'c', 'A': None}
```

amelie-rondot:
- fixes #1611 

---
This PR fixes incorrect errors which occur when validating data with missing required column, for the specific case of `schema_sync`option.

After [investigating](https://github.com/frictionlessdata/frictionless-py/issues/1611#issuecomment-1889655420), the better way to fix these incorrect errors we found is to remove missing columns from `field_info` `Row.__init__()` parameter at the creation of the `Row` in `row_stream`() local method.

Test cases have been added to traduce expected behavior:
Validation of tabular data with schema with one or many required fields, and data with one or many missing required columns and `schema_sync` option is set to `true`

Other test cases have not been broken.

Expected behavior could also be observed with command line:
```
frictionless run validate data.csv --schema schema.json --schema-sync
```
with:
```
schema.json
----
{
    "$schema": "https://frictionlessdata.io/schemas/table-schema.json",
    "fields": [
        {
            "name": "A",
            "title": "Field A",
            "type": "string",
            "constraints": {"required": true}
        },
        {"name": "B", "title": "Field B", "type": "string"},
        {"name": "C", "title": "Field C", "type": "string"}
    ]
}

```
```
data.csv
----
B, C
b, c
```

Only one error missing-label in the header's field "A" occurs in the validation report, as expected.


roll:
Thanks @amelie-rondot !

Can you please `make format`?

amelie-rondot:
@roll thanks a lot for reviewing!
I committed format fixes.

frictionless/resources/table.py
```diff
@@ -310,7 +310,18 @@ def row_stream():
             for row_number, cells in enumerated_content_stream:
                 self.stats.rows += 1
 
-                # Create row
+                # NB: missing required labels are not included in the
```
roll: Is it possible to move it from the row iteration cycle? Currently, this code will be executed for every row in a table

roll (commented):
Thanks!

I added a comment

frictionless/resources/table.py
amelie-rondot: Thanks!
Indeed, you're right, it's possible to move this block before initializing `row_stream` `TableResource` attribute.
I've just committed it. I also added some new test cases to insure that resulting behaviour corresponds to the expected one.

roll (approved):
Thanks a lot!

PR #1626, opened 2024-01-25
#1622: Fix issues with nested context manager calls

### LINKED ISSUE #1622: frictionless transform unhandled exception on 200 line csv, but not 197 line csv
# Overview

I am finding a very strange error when doing a transfrom (either in python code or via the command line tool). Depending on the size of the input file the transform succeeds fine, or throws an "I/O operation on closed file" exception.  The number of lines required to trigger it seems to vary, even by execution environment.

On a M1 Mac Mini it's currently 198 lines crashes, 197 lines passes.  On a gitpod instance (Ubuntu), it was around the same yesterday, but today is more like 150.  In our code version it can take 10k lines+.   But there is always a size above which this fails (and a size far short of e.g. `settings.FIELD_SIZE_LIMIT`).

## Example Command Line

```
% frictionless transform data/crash-transform/data.csv --pipeline data/crash-transform/pipeline.json
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ [step-error] Step is not valid: "cell_replace" raises "[step-error] Step is not valid: "table_normalize" raises "[source-error] The data source has not supported or has         │
│ inconsistent contents: I/O operation on closed file. " "                                                                                                                         │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

### Pipeline.json

```json
{
  "steps": [
    { "type": "table-normalize" },
    { "type": "cell-replace", "pattern": "BLANK", "replace": "" },
    { "type": "cell-replace", "pattern": "blank", "replace": "" },
    { "type": "cell-replace", "pattern": "NULL", "replace": "" },
    { "type": "cell-replace", "pattern": "null", "replace": "" },
    {
      "name": "NewSumField",
      "type": "field-add",
      "formula": "Field1 + Field2"
    },
    { "name": "NewConstantField", "type": "field-add", "value": "NewValue" }
  ]
}
```

### data.csv
```csv
Field1,Field2,Random1,Random2,Random3,Random4,Random5,Random6,Random7,Random8,Random9
0,0,BLANK,blank,NULL,null,5val0,6val0,7val0,8val0,9val0
1,10,1val1,2val1,3val1,4val1,5val1,6val1,7val1,8val1,9val1
2,20,1val2,2val2,3val2,4val2,5val2,6val2,7val2,8val2,9val2
3,30,BLANK,2val3,3val3,4val3,5val3,6val3,7val3,8val3,9val3
... extend as needed ...
```

### Sample files

[data.csv](https://github.com/frictionlessdata/frictionless-py/files/13975125/data.csv)
[pipeline.json](https://github.com/frictionlessdata/frictionless-py/files/13975127/pipeline.json)

richardt-engineb:
# Summary

- fixes #1622 

## Problem statement

The `Resource` class is also a [Context Manager](https://docs.python.org/3/reference/datamodel.html#context-managers).  That is, it implements the `__enter()__` and `__exit()__` methods to allow the use of `with Resource(...)` statements.

Prior to this PR, there was no limit on nesting `with` statements on the same `Resource`, but this caused problems because while the second `__enter()__` allowed the `Resource` to already be open, the first `__exit()__` would `close()` the Resource while the higher level context would expect it to still be open.

This would cause errors like "ValueError: I/O operation on closed file", or the iterator would appear to start from part way through a file rather than at the start of the file, and other similar behaviour depending on the exact locations of the nested functions.

This was made more complex because these `with` statements were often far removed from each other in the code, hidden behind iterators driven by generators, etc. They also could have different behaviour depending on number of rows read, the type of Resource (local file vs inline, etc.), the different steps in a pipeline, etc. etc.

All this meant that the problem was rare, hard to reduce down to an obvious reproduction case, and not realistic to expect developers to understand while developing new functionality.

## Solution

This PR prevents nested contexts being created by throwing an exception when the second, nested, `with` is attempted.  This means that code that risks these issues can be quickly identified and resolved during development.

The best way to resolve it is to use `Resource.to_copy()` to copy so that the nested `with` is acting on an independent view of the same Resource, which is likely what is intended in most cases anyway.

This PR also updates a number of the internal uses of `with` to work on a copy of the Resource they are passed so that they are independent of any external code and what it might have done with the Resource prior to the library methods being called.

## Breaking Change

This is technically a breaking change as any external code that was developed using nested `with` statements - possibly deliberately, but more likely unknowingly not falling into the error cases - will have to be updated to use `to_copy()` or similar.

However, the library functions have all been updated in a way that doesn't change their signature or their expected behaviour as documented by the unit tests. All pre-existing unit tests pass with no changes, and added unit tests for the specific updated behaviour do not require any unusual constructs.  It is still possible that some undocumented and untested side effect behaviours are different than before and any code relying on those may also be affected (e.g. `to_petl()` iterators are now independent rather than causing changes in each other)

So it is likely that very few actual impacts will occur in real world code, and the exception thrown does it's best to explain the issue and suggest resolutions.

# Tests

- All existing unit tests run and pass unchanged
- New unit tests were added to cover the updated behaviour 
-- These unit tests were confirmed to fail without the updates in this PR (where appropriate). 
-- These unit tests now pass with the updated code.
- The original script that identified the issue in #1622 was run and now gives the correct result (all rows appropriately converted and saved to file)

### DIFF
--- a/frictionless/analyzer/analyzer.py
+++ b/frictionless/analyzer/analyzer.py
@@ -34,7 +34,8 @@ class Analyzer:
         # Iterate rows
         columns_data: Dict[str, List[Any]] = {}
         numeric = ["integer", "numeric", "number"]
-        with resource:
+        # Use a copy of the resource to avoid side effects (see #1622)
+        with resource.to_copy() as resource:
             for row in resource.row_stream:
                 null_columns = 0
                 for field_name in row:
--- a/frictionless/formats/csv/parser.py
+++ b/frictionless/formats/csv/parser.py
@@ -63,7 +63,8 @@ class CsvParser(Parser):
             "wt", delete=False, encoding=self.resource.encoding, newline=""
         ) as file:
             writer = csv.writer(file, **options)  # type: ignore
-            with source:
+            # Use a copy of the source to avoid side effects (see #1622)
+            with source.to_copy() as source:
                 if self.resource.dialect.header:
                     writer.writerow(source.schema.field_names)
                 for row in source.row_stream:
--- a/frictionless/formats/excel/parsers/xls.py
+++ b/frictionless/formats/excel/parsers/xls.py
@@ -109,7 +109,8 @@ class XlsParser(Parser):
         if isinstance(title, int):
             title = f"Sheet {control.sheet}"
         sheet = book.add_sheet(title)
-        with source:
+        # Write from a copy of the source to avoid side effects (see #1622)
+        with source.to_copy() as source:
             if self.resource.dialect.header:
                 for field_index, name in enumerate(source.schema.field_names):
                     sheet.write(0, field_index, name)
--- a/frictionless/formats/excel/parsers/xlsx.py
+++ b/frictionless/formats/excel/parsers/xlsx.py
@@ -148,7 +148,8 @@ class XlsxParser(Parser):
         if isinstance(title, int):
             title = f"Sheet {control.sheet}"
         sheet = book.create_sheet(title)
-        with source:
+        # Write from a copy of the source to avoid side effects (see #1622)
+        with source.to_copy() as source:
             if self.resource.dialect.header:
                 sheet.append(source.schema.field_names)
             for row in source.row_stream:
--- a/frictionless/formats/gsheets/parser.py
+++ b/frictionless/formats/gsheets/parser.py
@@ -53,7 +53,8 @@ class GsheetsParser(Parser):
         sh = gc.open_by_key(key)
         wks = sh.worksheet_by_id(gid) if gid else sh[0]  # type: ignore
         data: List[Any] = []
-        with source:
+        # Use a copy of the source to avoid side effects (see #1622)
+        with source.to_copy() as source:
             data.append(source.schema.field_names)
             for row in source.row_stream:
                 data.append(row.to_list())
--- a/frictionless/formats/html/parser.py
+++ b/frictionless/formats/html/parser.py
@@ -57,7 +57,8 @@ class HtmlParser(Parser):
     # It will give us an ability to support HtmlDialect
     def write_row_stream(self, source: TableResource):
         html = "<html><body><table>\n"
-        with source:
+        # Use a copy of the source to avoid side effects (see #1622)
+        with source.to_copy() as source:
             html += "<tr>"
             for name in source.schema.field_names:
                 html += f"<td>{name}</td>"
--- a/frictionless/formats/inline/parser.py
+++ b/frictionless/formats/inline/parser.py
@@ -91,7 +91,8 @@ class InlineParser(Parser):
     def write_row_stream(self, source: TableResource):
         data: List[Any] = []
         control = InlineControl.from_dialect(self.resource.dialect)
-        with source:
+        # Write from a copy of the source to avoid side effects (see #1622)
+        with source.to_copy() as source:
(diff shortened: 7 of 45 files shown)


roll:
Thanks @richardt-engineb !

It's a complex change I will ask for a peer review in the chat

augusto-herrmann (commented):
I have never used nested `with` statements with `Resource` objects, so I do not think this would directly affect our own existing code.

However, by looking the code on this PR I can notice that a lot of internal code of the Frictionless Framework was changed to use `Resource.to_copy()` when using context managers. Does copying a Resource object entail copying the data as well or just metadata? If this does indeed create an in-memory copy of the whole resource, including data, this is likely to increase memory usage, which might be an issue if the Resource includes a lot of data.

It would be nice to make a test comparing memory usage and running time before and after the change, especially with large Resources.

richardt-engineb:
Thanks for the commens @augusto-herrmann (and sorry about the delay in replying)

> I have never used nested `with` statements with `Resource` objects, so I do not think this would directly affect our own existing code.

It's good to hear that it wouldn't affect your own use of Frictionless.  The main cases that would affect would be where things in the library happen to be called in such a way that nested `with` statements happen.  E.g. when using a pipeline with steps, if they re-uses the same resource object the whole way through (and not copies) you can end up with nested `with` statements (usually related to `to_petl()`-based generators etc.)

> However, by looking the code on this PR I can notice that a lot of internal code of the Frictionless Framework was changed to use `Resource.to_copy()` when using context managers. Does copying a Resource object entail copying the data as well or just metadata? If this does indeed create an in-memory copy of the whole resource, including data, this is likely to increase memory usage, which might be an issue if the Resource includes a lot of data.
> 
> It would be nice to make a test comparing memory usage and running time before and after the change, especially with large Resources.

`to_copy()` is:

```python
def to_copy(self, **options: Any) -> Self:
        """Create a copy from the resource"""
        return super().to_copy(
            data=self.data,
            basepath=self.basepath,
            detector=self.detector,
            package=self.package,
            **options,
        )
```

The main thing in there that could take up space is indeed the `data` which could noticeably increase the space used.  However the mitigation would be that if using very large datasets it seems likely that data would come from loading a file which is only using generators/iterators rather than caching the whole file in memory.  And conversely, anywhere that full in-memory `data` is being used is likely to be small enough that the memory increase is not a concern overall.

Are there any tests you would specifically like to see to increase your confidence, or use-cases where huge in-memory data sets are be used in ways that this would make harder?

Even in this change, someone with huge in-memory data-sets could use `open` and `close` explicitly to avoid copies.  That wouldn't affect the internals however, so it does become a question of correctness vs memory use.  I did consider doing reference counting instead of copying, however that still had the problem where the state of the generators/iterators was inconsistent depending on what else had been called (e.g. pipeline ordering, etc).  So e.g. the `tests/table/test_to_petl.py` tests that I added _do not_ pass without the `to_copy()` changes I made (and would not pass in a reference counting approach).

I also considered adding a `SAFE_ITERATORS` or similar flag that would either copy or not copy depending on value, but that also seemed like asking for trouble in the future as when the issues happen they are very subtle and hard to understand.  So it would almost need to be a `DANGEROUS_THINGS_WILL_EVENTUALLY_GO_WRONG` flag or something!

As I said, thanks for the feedback, and happy to look at any additional tests etc. that you think would conclusively show a meaningful difference that would convince you either way.

PR #1627, opened 2024-01-25
chore: fix minor typos

afuetterer:
This PR fixes some minor typos, no issue created, because this should be trivial.

These changes were done with codespell.

It also wanted to replace "underlaying" -> "underlying". Which I ignored. That is correct, right?


roll (approved):
Thanks!

PR #1674, opened 2024-08-23
Fix deprecated lazy_fixture imports

pdelboca:
This PR fixes several deprecations that were making our tests fail.

I'm also removing `pytest-only` dependency since it does not support Python 3.12 and we are not currently using it.

---


pierrecamilleri (approved):
:pray:

PR #1683, opened 2024-09-13
fix: Pandas parser does fail to parse integer or boolean only dataframes

### LINKED ISSUE #1678: reading data from pandas returns null values
following the docs: 
https://framework.frictionlessdata.io/docs/formats/pandas.html

to reproduce: 

```console
mamba create -n dp-test python pandas -y
mamba activate dp-test 
pip install frictionless frictionless[pandas]
```

```py
# >> python
# Python 3.12.5 | packaged by conda-forge | (main, Aug  8 2024, 18:36:51) [GCC 12.4.0] on linux
# Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas as pd
>>> x = range(0, 5)
>>> y = [_**2 for _ in x]
>>> df = pd.DataFrame({"x": x, "y": y})
>>> from pprint import pprint
>>> pprint(df)
   x   y
0  0   0
1  1   1
2  2   4
3  3   9
4  4  16
>>> from frictionless import Resource
>>> r = Resource(df)
>>> r.read_rows()
[{'x': None, 'y': None}, {'x': None, 'y': None}, {'x': None, 'y': None}, {'x': None, 'y': None}, {'x': None, 'y': Non
```

pierrecamilleri:
- fixes #1678 

Converting the Series returned by `iterrows()` to a dict converts `np.int64` type to python's native `int` type and fixes the bug (same with booleans). 

+ Adding non-regression tests

- I was also concerned with the next lines, especially : `if value is np.nan: value = None` 
  - it was untested, so I added a test. It looks like `to_dict` would not change the behavior of `np.nan` conversion (see side note), so I left this code unchanged.
- Primary keys are returned as `int`s or `tuple[int]`, no `np.int64` there
- `Timestamps` types are kept unchanged, so the `if isinstance(value, pd.Timestamp):` still applies.

## Side note

`np.nan` behavior is quite strange with `df.iterrows()` : in a number column, it will be converted to `float("nan")`, whereas in string column it will be kept as `np.nan`.  Adding `to_dict()` to the row Series does not change the types.

### DIFF
--- a/frictionless/formats/pandas/__spec__/test_parser.py
+++ b/frictionless/formats/pandas/__spec__/test_parser.py
@@ -2,6 +2,7 @@ from datetime import datetime, time
 from decimal import Decimal
 
 import isodate
+import numpy as np
 import pandas as pd
 import pytz
 from dateutil.tz import tzoffset, tzutc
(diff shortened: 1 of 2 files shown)


pierrecamilleri:
@roll @pdelboca ready for review !

pdelboca (approved):
Looks good! I like changes followed by tests 👍🏼

PR #1690, opened 2024-09-23
fix: do not detect resource type if type is provided by user

### LINKED ISSUE #1688: Schema descriptor with `path` property raises a "resource-error"
# Overview

If a schema descriptor has a `path` property, validating the descriptor with python code returns a "resource-error". At the command line, the file is considered valid.

For the context, the `path` property is defined for a [data resource](https://datapackage.org/standard/data-resource/#tabular-path-or-data), but *not* in the Table Schema specification. It is also not in the [Schema class properties](https://github.com/frictionlessdata/frictionless-py/blob/v5.17.1/frictionless/schema/schema.py).

I have no clue what is going on here. 

# Reproduce 

Schema descriptor in file "schema.json" :

```json
{
  "name": "schema",
  "path": "abc",
  "fields": [
    {
      "name": "id",
      "type": "integer"
    }
  ]
}
```

```python
import frictionless

schema = frictionless.Schema("./schema.json")
descr = schema.to_descriptor()
print(frictionless.validate(descr, type="schema"))
# {'valid': False,
#  'stats': {'tasks': 0, 'errors': 1, 'warnings': 0, 'seconds': 0},
#  'warnings': [],
#  'errors': [{'type': 'resource-error',
#              'title': 'Resource Error',
#              'description': 'A validation cannot be processed.',
#              'message': "The data resource has an error: [{'name': 'id', "
#                         "'type': 'integer'}] is not of type 'integer' at "
#                         "property 'fields'",
#              'tags': [],
#              'note': "[{'name': 'id', 'type': 'integer'}] is not of type "
#                      "'integer' at property 'fields'"}],
#  'tasks': []}
```

# What I tested

- Validating at the command line  `frictionless validate --type=schema --json schema.json` returns `VALID`
- Validating with file path returns `'valid': True`
- Changing the name of `path` to `patch` does not reproduce the behavior. 
- Changing the `path` value to an URL does not solve the issue.

pierrecamilleri:
- fixes #1688


roll (approved):
Thanks!

PR #1691, opened 2024-09-26
feat: indexer option to skip constraints creation

roll:
- Open Data Editor needs a way to create [flexible database tables](https://github.com/okfn/opendataeditor/issues/552) that can have some constraint violations for further fixing. So here is a new `Indexer.without_constraints` flag to achieve this goal. 

---

@pierrecamilleri 
can you please take a look?

### DIFF
--- a/frictionless/formats/sql/adapter.py
+++ b/frictionless/formats/sql/adapter.py
@@ -122,6 +122,7 @@ class SqlAdapter(Adapter):
         table_name: str,
         force: bool = False,
         with_metadata: bool = False,
+        without_constraints: bool = False,
     ) -> None:
         with self.engine.begin() as conn:
             if force:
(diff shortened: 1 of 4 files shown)


frictionless/formats/sql/adapter.py
```diff
@@ -122,6 +122,7 @@ def write_schema(
         table_name: str,
         force: bool = False,
         with_metadata: bool = False,
+        without_constraints: bool = False,
```
pierrecamilleri: nitpick : I think `ignore_constraints` would be a little bit more explicit

frictionless/formats/sql/mapper.py
```diff
@@ -252,15 +269,20 @@ def write_field(self, field: Field, *, table_name: str) -> Column:  # type: igno
...
-        column_kwargs = {"nullable": nullable, "unique": unique}
+        column_args = [field.name, column_type]  # type: ignore
+        column_kwargs = {}
         if field.description:
             column_kwargs["comment"] = field.description
-        column = sa.Column(*column_args, **column_kwargs)
+        if not without_constraints:
+            column_args += checks  # type: ignore
+            column_kwargs["nullable"] = nullable
+            column_kwargs["unique"] = unique
```
pierrecamilleri: Nitpick: For better readability, maybe I would bring the definitions of checks, nullable and unique in this block (under the condition `if not without_constraints`). This has the advantage of dealing with constraints at a single place and not assigning variables that are not used.

pierrecamilleri (approved):
Looks good to me !

frictionless/formats/sql/adapter.py
roll: Thanks! Renamed to `ignore_constraints`

roll:
Thanks!

PR #1722, opened 2024-12-13
fix: parallel file validation for a datapackage

### LINKED ISSUE #1644: Adding --parallel CLI flag breaks validation
# Overview

Adding `--parallel` to `frictionless validate` breaks validation and fails. I could really use this flag because multiple files in this dataset are ≥ 2GB, some approaching 7GB.

# Details

My package passes validation by running this command:

```
frictionless validate --trusted package.json
```

However, if I add `--parallel` it fails immediately:

```
/tmp # frictionless validate --trusted --parallel package.json
╭─ Error ───────────────────────────────────────────────────────╮
│ name 'Resource' is not defined                                │
╰───────────────────────────────────────────────────────────────╯
```

I have the following package definition file, here formatted as YAML instead of JSON for readability:

```yaml
resources:
  - encoding: utf-8
    format: csv
    mediatype: text/csv
    name: medical_observations
    path: /var/dataset/output/Observations_20230917102633.csv
    schema: /app/schemas/medical/observations.yaml
    type: table
  - encoding: utf-8
    format: csv
    mediatype: text/csv
    name: medical_medications
    path: /var/dataset/output/Medications_20230917102633.csv
    schema: /app/schemas/medical/medications.yaml
    type: table
  - encoding: utf-8
    format: csv
    mediatype: text/csv
    name: medical_vitals
    path: /var/dataset/output/Vitals_20230917102633.csv
    schema: /app/schemas/medical/vitals.yaml
    type: table
  - encoding: utf-8
    format: csv
    mediatype: text/csv
    name: medical_patient
    path: /var/dataset/output/Patient_20230917102632.csv
    schema: /app/schemas/medical/patient.yaml
    type: table
  - encoding: utf-8
    format: csv
    mediatype: text/csv
    name: medical_toxicity
    path: /var/dataset/output/Toxicity_20230917102635.csv
    schema: /app/schemas/medical/toxicity.yaml
    type: table
  - encoding: utf-8
    format: csv
    mediatype: text/csv
    name: medical_encounter
    path: /var/dataset/output/Encounter_20230917102627.csv
    schema: /app/schemas/medical/encounter.yaml
    type: table
  - encoding: utf-8
    format: csv
    mediatype: text/csv
    name: medical_problem
    path: /var/dataset/output/Problem_20230917102634.csv
    schema: /app/schemas/medical/problem.yaml
    type: table
```

# Environment

This is running on the official Docker image for Python on Alpine 3.19; Python versions tried are 3.12.2,  3.11.7, and 3.10.13.

`uname -a`:
```
Linux 3ff4bcba25f0 6.6.12-linuxkit #1 SMP Fri Jan 19 08:53:17 UTC 2024 aarch64 Linux
```

`frictionless --version`:
```
5.16.1
```

`pip3 freeze`:
```
annotated-types==0.6.0
attrs==23.2.0
certifi==2024.2.2
chardet==5.2.0
charset-normalizer==3.3.2
click==8.1.7
colorama==0.4.6
frictionless==5.16.1
humanize==4.9.0
idna==3.6
isodate==0.6.1
Jinja2==3.1.3
jsonschema==4.17.3
markdown-it-py==3.0.0
marko==2.0.2
MarkupSafe==2.1.5
mdurl==0.1.2
petl==1.7.14
pydantic==2.6.1
pydantic_core==2.16.2
Pygments==2.17.2
pyrsistent==0.20.0
python-dateutil==2.8.2
python-slugify==8.0.4
PyYAML==6.0.1
requests==2.31.0
rfc3986==2.0.0
rich==13.7.0
setuptools==69.1.0
shellingham==1.5.4
simpleeval==0.9.13
six==1.16.0
stringcase==1.2.0
tabulate==0.9.0
text-unidecode==1.3
typer==0.9.0
typing_extensions==4.9.0
urllib3==2.2.1
validators==0.22.0
wheel==0.42.0

```

pierrecamilleri:
- fixes #1644 

### Context

- In validator.py, `Resource` was imported as a type, but then used as a class to retrieve a class method (`Resource.from_descriptor`).
- The tests would not intercept this error, as they were performed on a schema with a foreign key, and the validation would fall back on synchronous validation in this case.   
- Importing `Resource` (not as a type) into `validator.py` leads to a circular import. A small workaround would have been to import Resource locally, but this was not very satisfactory. 

### Suggested changes and fixes

- I repaired the tests, using datapackages with no foreign keys to test parallel validation. I checked that these tests would have failed because with the bug. 
- I moved the implementation of `Validator.validate_resource` and `Validator.validate_package` methods to the `Resource.validate` and `Package.validate` methods. This solves the circular import problem, and removes duplication (method signatures). 
- The tests have been dispatched as well.
- I soft deprecated the `Validator` class, keeping the methods (forwarding input parameters) as they are part of the public API. No warning, only a deprecation message in the docstring, and a mention that there is no plan to remove the class in the future.

### Note 

I could not produce any significant performance gain in my tests, I wonder if there is some shared resources' lock that prevents the validation from effectively running the validation in parallel (but I have not spent much time on this yet).

<details> 
<summary>Archive of explorations</summary>

- [X] Test that it works correctly
  - Works without error, however,  I could not produce any significant performance gain in my tests.
- [X] Check why the tests did not intercept this error
  - The tests for parallel processing are run for a schema with a foreign key - which fall backs to single-core processing. So the `validate_parallel` function was actually never executed in the tests.  
- [X] See if I can find a more elegant way to deal with the circular import. I actually don't see the benefit of having this validator.py class instead of dispatching the methods to `Resource` and `DataPackage` directly, but I may have missed something.
</details>


frictionless/resource/resource.py
```diff
@@ -619,8 +619,112 @@ def validate(
...
+
+        # Validate metadata
+        try:
+            self.to_descriptor(validate=True)
+        except FrictionlessException as exception:
+            return Report.from_validation_task(
+                self, time=timer.time, errors=exception.to_errors()
+            )
+
+        # TODO: remove in next version
```
pdelboca: I usually like to add more explicit messages for future maintainers. 

Something like:
- `Remove in X.Y.Z when foo gets deprecated`
- `Remove when Y is removed`
- `Remove when we no longer support X`

pdelboca (approved):
Thanks for this @pierrecamilleri !

Just one small comment but feel free to merge :)

frictionless/resource/resource.py
pierrecamilleri: Yes indeed ! (This was copy-pasted). 
I think this is a breaking change from v4 to v5, so let's say  "remove in v6".
(I'll check for any dependency to the hash, but to me this feature is dropped with no dependencies)

PR #1729, opened 2025-01-22
#1728: add pass_row as configurable parameter for field_update

### LINKED ISSUE #1728: pass_row parameter should be an input to the field_update step to support functions
# Summary

In the `field_update` step, the `pass_row` parameter is set to `False` by default, except for `formula` which sets it to `True` before calling the `table.convert()` from petl. We should allow this parameter to be set to `True` for `functions` as well for the same reasons that `formula` sets it to `True`.

ebAbhay:
# Description
As described in #1728, pass_row parameter is now added as an editable parameter for the field_update step. This allows us to use row data inside custom python function.

The default value of the parameter remains False so that any existing uses of this functionality will not be affected.

# Tests
Ensured the newly added unit test ` test_step_field_update_with_function_and_pass_row_true` runs without fail and also that all the previous test cases run and pass too.

- fixes #1728

### DIFF
--- a/frictionless/steps/field/field_update.py
+++ b/frictionless/steps/field/field_update.py
@@ -50,11 +50,16 @@ class field_update(Step):
     A descriptor for the field to set the metadata.
     """
 
+    pass_row: bool = False
+    """
+    Pass the entire row as a parameter for the transformation function when True.
+    """
+
     # Transform
 
     def transform_resource(self, resource: Resource):
         function = self.function
-        pass_row = False
+        pass_row = self.pass_row
         table = resource.to_petl()  # type: ignore
         descriptor = deepcopy(self.descriptor) or {}
         new_name = descriptor.get("name")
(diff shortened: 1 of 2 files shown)


pierrecamilleri:
Thanks for the PR ! I'll try to have a look at it soon.

frictionless/steps/field/field_update.py
```diff
@@ -50,11 +50,16 @@ class field_update(Step):
     A descriptor for the field to set the metadata.
     """
 
+    pass_row: bool = False
+    """
+    Pass the entire row as a parameter for the transformation function when True.
```
pierrecamilleri: I understand from the test that the row parameter is passed _in addition_ to the value, which is not obvious in this description.

I suggest to remove this ambiguity.

```suggestion
    Pass the entire row as a second parameter for the transformation function if set to True.
    The first parameter (the value) remains unchanged.
```

pierrecamilleri (approved):

frictionless/steps/field/field_update.py
ebAbhay: Thank you for the suggestion, it is a valid change and I have committed it.

pierrecamilleri:
Thanks a lot for your contribution!

PR #1736, opened 2025-02-07
Fix documentation typos

paulkorir:
A number of typos in the documentation were incorrect resulting on poor user experience. In particular, the link to the types of steps was to a Markdown file instead of HTML.

### DIFF
--- a/.github/workflows/general.yaml
+++ b/.github/workflows/general.yaml
@@ -101,7 +101,7 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-      - name: Install Python
+      - name: Install Pythonha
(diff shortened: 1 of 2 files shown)


.github/workflows/general.yaml
```diff
@@ -101,7 +101,7 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-      - name: Install Python
+      - name: Install Pythonha
```
pierrecamilleri: This seems to be unintended, or is it ?

pierrecamilleri (approved):

pierrecamilleri:
Thanks !

PR #1758, opened 2025-10-10
use an authentic User-Agent

jze:
Use `frictionless-py` + version as the User-Agent instead of a faked one.

- fixes#1757


pierrecamilleri:
Thank you @jze  ! I have to look into why the CI breaks before merging, probably an obsolete dependency.

pierrecamilleri (approved):

augusto-herrmann (approved):
I agree with this change. It's always best to be honest identifying the User Agent. Too many blocks nowadays, especially with the generalized backlash against the rise of the AI crawlers.

PR #1759, opened 2025-12-05
refactor: move IntegerField to pydantic

BmnQuentin:
Integrated pydantic logic to IntegerField.
All tests pass (apart from 5 console ones not related to modifications)
Fixed a few issued:
- examples should enable Any type, not only strings
- descriptor was broken during the merge of two kinds

### DIFF
--- a/frictionless/fields/field_descriptor.py
+++ b/frictionless/fields/field_descriptor.py
@@ -1,16 +1,11 @@
-"""field_descriptor.py provides pydantic Models for Field descriptors"""
-
 from __future__ import annotations
 
 import datetime
-from typing import Any, ClassVar, Dict, List, Literal, Optional, Union, Pattern
-import re
+from typing import List, Literal, Optional, Union
 
-from pydantic import Field as PydanticField, AliasChoices, model_validator, BaseModel
-from typing_extensions import Self
+from pydantic import Field as PydanticField, BaseModel
 
-
-from .. import settings
+from .base_field_descriptor import BaseFieldDescriptor
 from .field_constraints import (
     BaseConstraints,
     CollectionConstraints,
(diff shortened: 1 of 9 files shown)


BmnQuentin:
TODO: if the global idea is ok, move each class from `field_descriptor` in its own file (eg `boolean.py`)

frictionless/fields/field_descriptor.py
```diff
@@ -122,18 +136,6 @@ def write_value(self, cell: Optional[bool]) -> Optional[str]:
             return self.true_values[0] if cell else self.false_values[0]
         return None
 
-    @pydantic.model_validator(mode="after")
```
pierrecamilleri, line 125: Why is this check removed ?

frictionless/fields/field_descriptor.py
```diff
@@ -225,24 +227,85 @@ class IntegerFieldDescriptor(BaseFieldDescriptor):
...
+
+        elif isinstance(cell, int):
+            return cell
+
+        elif isinstance(cell, str):
+            cell = cell.strip()
+
+            # Process the cell (remove non-digit characters if bare_number is False)
+            if not self.bare_number:
+                pattern = re.compile(r"((^[^-\d]*)|(\D*$))")
```
pierrecamilleri: Read value will be used once for each row, for performance reasons it would be useful to compile the regex outside of the function.

frictionless/schema/field.py
```diff
@@ -271,15 +279,29 @@ def metadata_import(
...
+            # Set by_alias=True to get camelCase keys used by Frictionless (bareNumber) instead of snake_case (bare_number)
+            # Exclude 'name' from descriptor_descr because it may be "shared" (coming from detector.py)
+            descriptor_descr = self._descriptor.model_dump(
+                exclude_none=True, exclude_unset=True, by_alias=True, exclude={"name"}
+            )
             ## Temporarily, Field properties have priority over
             ## Field._descriptor properties
-            descr = {**descr, **super().to_descriptor(validate=validate)}
+            ## Merge descriptor_descr into base_descr to preserve base order
+            descr = {**base_descr, **descriptor_descr}
```
pierrecamilleri, line 313: If changing the order, the above comment needs to be updated ("_descriptor" has priority)

pierrecamilleri (approved):
> (apart from 5 console ones not related to modifications)

Were they already failing before ?  Even on master branch ? 
(normaly CI was passing not so long ago, however I have not investigated CI breaking in https://github.com/frictionlessdata/frictionless-py/pull/1758 yet)

frictionless/fields/field_descriptor.py
BmnQuentin: It was moved to a more global logic (see lines 66->79)

pierrecamilleri: Ah sorry, thought it was IntegerField specific, all good ;)

BmnQuentin:
Changes in #1751 already break console tests. Will try and fix that in another PR

EDIT: fixed in #1752, needs rebase

frictionless/schema/field.py
BmnQuentin: As i understand it, it is already the correct way?

pierrecamilleri: I mean this comment :

>   Temporarily, Field properties have priority over
>   Field._descriptor properties

It seems the opposite to me (descriptor_descr overwrites / has priority over base_descr)

PR #1777, opened 2026-05-04
Support csv format for ArrayField

### LINKED ISSUE #1434: Support `csv` format for `ArrayField`
# Overview

Currently, it's only possible to express an array within e.g. a csv this way:

```
type: array
format: default
```
```
["v1", "v2", "v3"]  # note that it should be additionally quoted within csv
```

We'd like to support `csv` format to allow this:

```
type: array
format: csv
```
```
v1, v2, v3  # note that it should be additionally quoted within csv
```

# Implementation

It's relatively simple to implement by updating `fields.ArrayField` and using builtin `csv` module of Python.

- [ ] write `csv_cell_reader` function in `arrayField.create_cell_reader`
- [ ] based on `field.format: default | csv` return `cell_reader` or `csv_cell_reader`
- [ ] add tests
- [ ] add docs

# References

- https://github.com/frictionlessdata/specs/issues/736.

ejscott1:
- fixes #1434

---

## Summary

- Add `format="csv"` support for `ArrayField`
- Parse CSV array values with Python's built-in `csv` module
- Add tests for CSV values, quoted commas, spaced delimiters, invalid values, and `array_item` conversion
- Add an ArrayField docs example for CSV format

## Tests

- `hatch run python -m pytest frictionless\fields\__spec__\test_array.py -q`

### DIFF
--- a/docs/fields/array.md
+++ b/docs/fields/array.md
@@ -14,6 +14,16 @@ rows = extract(data, schema=Schema(fields=[fields.ArrayField(name='name')]))
 print(rows)
 ```
 
+Use `format="csv"` to read comma-separated array values.
+
+```python script tabs=Python
+from frictionless import Schema, extract, fields
+
+data = [["name"], ["value1, value2"]]
+rows = extract(data, schema=Schema(fields=[fields.ArrayField(name="name", format="csv")]))
+print(rows)
+```
+
 ## Reference
 
 ```yaml reference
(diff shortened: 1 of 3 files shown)


pierrecamilleri:
As mentioned in the issue, the array "format" property has been abandoned in favor of the datapackage v2 "list" field type.

PR #1783, opened 2026-06-03
feat: accept DPv2 missing values object with labels.  

pierrecamilleri:
https://datapackage.org/overview/changelog/#schemamissingvalues-updated

# TODO

- `MissingValues` class is currently a breaking change 
- Validation is always that of DP2 even for DP1 schema


pierrecamilleri:
Superseded by #1787 that has no breaking change.

PR #1784, opened 2026-06-08
feat: accept `$schema` property

pierrecamilleri:
Adds `$schema` reading from descriptor and writing to_descriptor for datapackage v2 : 
- package.$schema
- resource.$schema
- schema.$schema
- dialect.$schema

> A root level [xxx] descriptor MAY have a $schema property

No validation is performed against the $schema properties yet.

### implementation details

Implemented on Metadata (`_schema_profile` property, to not clash with e.g. `resource.schema` and as `$schema` is not a valid python symbol), but only child classes with explicit declaration will read it (for others, it will be considered a `custom` property).

### DIFF
--- a/frictionless/dialect/dialect.py
+++ b/frictionless/dialect/dialect.py
@@ -26,6 +26,12 @@ class Dialect(Metadata, metaclass=Factory):
     # TODO: add docs
     """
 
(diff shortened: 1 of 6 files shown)


zaneselvans:
Thanks for adding this!  I recently found that I had to manually add this to some new v2 datapackages I was writing out and was very surprised.

PR #1786, opened 2026-06-11
feat: infer datapackage version from $schema and propagate via inheritence

pierrecamilleri:
Adds the `datapackage_version` property.


pierrecamilleri:
Oopsie this branch deviated from its initial purpose, which was to infer datapackage version for validation and parsing, and then forget about it. Instead I'v transformed it into a permanent state that is best avoided. Closing for another attempt.

PR #1799, opened 2026-07-28
fix: avoid cell errors for header mismatches

### LINKED ISSUE #1791: bug: missing field generates "missing-cell" errors additionnaly to the "missing-label" error
# Description

When providing data with missing fields/labels compared to the schema fields, and provided that the validation happens without "--schema-sync" (or soon, with default "fieldsMatch": "exact" property), a "missing-label" error is raised (expected), but also many "missing-cell" errors (unexected). 

# Expected behavior 

There should not be a "missing-cell" error in the report, for two reasons : 

- it does not add any useful information to the "missing-label" error
- it's use should be limited to irregular rows (e.g. missing value in a csv row), not for a missing field in the data.    

# To reproduce 

```data.csv
a
1
2
```

```dp.json
{
  "resources": [
    {
      "name": "data",
      "path": "data.csv",
      "format": "csv",
      "schema": {
        "fields": [
          {"name": "a"},
          {"name": "b"}
          ]
      }
    }
  ]
}
```

```bash
frictionless validate dp.json
```

EDIT: same problem with extra-cell ! (and same origin)

stefan1condeescu:
- fixes #1791

With `fieldsMatch: "exact"`, a header mismatch was reported twice: once as `missing-label` or `extra-label`, and again as a cell error for every row.

This PR makes the expected fields follow the header length. Integrity checks and analyzer statistics now only use fields that are actually present, while irregular rows still report cell errors as expected.

I added regression tests for missing and extra labels, integrity constraints, and irregular rows. The full test suite passes (`1876 passed, 292 skipped`).

### DIFF
--- a/frictionless/table/header.py
+++ b/frictionless/table/header.py
@@ -140,7 +140,8 @@ class Header(List[str]):  # type: ignore
     def get_expected_fields(self) -> List[Field]:
         """Returns the fields, in the order expected in the data.
 
-        Under `exact`, this is just the schema fields unchanged.
+        Under `exact`, schema fields keep their order and are truncated or
+        extended with `any`-typed fields to match the labels.
 
         Under the name-matched modes, fields are reordered to match the labels;
         labels without a matching field get a fresh `any`-typed field (even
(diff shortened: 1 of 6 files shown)


stefan1condeescu:
Hi @pierrecamilleri, just following up on this PR when you get a chance. All checks are passing. Thanks!

frictionless/table/header.py
```diff
@@ -151,10 +152,28 @@ def get_expected_fields(self) -> List[Field]:
...
+
+            for label in self.__labels[len(expected) :]:
+                name = label
+                suffix = 2
+
+                while name in used_names:
+                    name = f"{label}{suffix}"
+                    suffix += 1
+
+                used_names.add(name)
```
pierrecamilleri, line 171: Thanks for the label deduplication. 

Unicity of "fields" is indeed expected in the codebase (e.g. [Row indexes fields by their name](https://github.com/frictionlessdata/frictionless-py/blob/44661adf0c9ec593d48642ba380ce33bbad95e15/frictionless/table/row.py#L35)). 

There is however another (actually, incorrect) deduplication logic [in the Detector class](https://github.com/frictionlessdata/frictionless-py/blob/main/frictionless/detector/detector.py#L344), the behavior should be harmonized and factorized.

frictionless/table/header.py
```diff
@@ -151,10 +152,28 @@ def get_expected_fields(self) -> List[Field]:
...
+                expected.append(Field.from_descriptor({"name": name, "type": "any"}))
```
pierrecamilleri, line 172: There may be a factorizing opportunity with the branch matching by names : for missing fields, add an "any" field. What differs is the way "missing fields" are identified.

pierrecamilleri (approved):
Thank you! 
I noted some factorizing opportunities, however they likely go beyond the scope of this fix, I will deal with them in a separate PR.

Issue #1737, opened 2025-02-11
Implement changes related to datapackage version 2


pierrecamilleri:
This issue gives information to track the adaptation of frictionless-py to the version 2 of the datapackage specification. 

Right now, no single feature of datapackage v2 has been implemented in frictionless-py , but hopefully we will be working on this soon. I'll keep this issue up to date as soon as the developments become more concrete.

- [ ] [Tabular Data Package (removed)](https://datapackage.org/overview/changelog/#tabular-data-package-removed)
- [ ] [package.$schema (new)](https://datapackage.org/overview/changelog/#packageschema-new)
- [ ] [package.contributors (updated)](https://datapackage.org/overview/changelog/#packagecontributors-updated)
- [ ] [package.version (updated)](https://datapackage.org/overview/changelog/#packageversion-updated)
- [ ] [package.sources (updated)](https://datapackage.org/overview/changelog/#packagesources-updated)
- [ ] [resource.name (updated)](https://datapackage.org/overview/changelog/#resourcename-updated)
- [ ] [resource.path (updated)](https://datapackage.org/overview/changelog/#resourcepath-updated)
- [ ] [resource.type (new)](https://datapackage.org/overview/changelog/#resourcetype-new)
- [ ] [resource.$schema (new)](https://datapackage.org/overview/changelog/#resourceschema-new)
- [ ] [resource.encoding (updated)](https://datapackage.org/overview/changelog/#resourceencoding-updated)
- [ ] [resource.sources (updated)](https://datapackage.org/overview/changelog/#resourcesources-updated)
- [ ] [dialect.schema (new)](https://datapackage.org/overview/changelog/#dialectschema-new)
- [ ] [dialect.table (new)](https://datapackage.org/overview/changelog/#dialecttable-new)
- [ ] [schema.$schema (new)](https://datapackage.org/overview/changelog/#schemaschema-new)
- [ ] [schema.fieldsMatch (new)](https://datapackage.org/overview/changelog/#schemafieldsmatch-new)
- [ ] [schema.missingValues (updated)](https://datapackage.org/overview/changelog/#schemamissingvalues-updated)
- [ ] [schema.primaryKey (updated)](https://datapackage.org/overview/changelog/#schemaprimarykey-updated)
- [ ] [schema.uniqueKeys (new)](https://datapackage.org/overview/changelog/#schemauniquekeys-new)
[...]

Issue #1718, opened 2024-12-11
enum constraint does not validate array items for type: "array"


pierrecamilleri:
With a schema specifying an enum constraint on a field of type array, the enum constraints does not validate the array items but the whole array value.

# Expected behavior

Array items should be validated against the enum constraint.

# Data to reproduce 

schema.json
```json
{
  "fields": [
    {
      "name": "foo",
      "type": "array",
      "constraints": {
        "enum": ["A", "B"]
      }
    }
  ]
}
```

data.csv 
```csv
foo
"[""1"",""2""]"
"[""A"",""B""]"
```

```sh
frictionless --version
 # 5.18.0
```

pierrecamilleri:
Actually, after rereading the spec for array and enum, there is no hint to the enum constraint being supported in this way in the table schema v2 spec.

PR #1805, opened 2026-08-18
feat: list field type (DPv2)

### LINKED ISSUE #1741: Feature Request: Support for Multiple Occurrences in a Single Field
**Description**
Frictionless currently does not support validating multiple occurrences of values within a single field using built-in constraints. We have a use case where a field can contain multiple values from a predefined list, separated by semicolons (;). However, enforcing this pattern within the schema requires a workaround using regex patterns.

**Use Case**
We need to validate fields like OBSERVATION_CATEGORY_SDOH_CODE and OBSERVATION_CATEGORY_SDOH_TEXT, where values come from a fixed list but can appear multiple times in a single field, separated by semicolons.

✅ Example of valid values:

"food-insecurity"
"food-insecurity; housing-instability; employment-status"
"stress; social-connection" 

**Proposed Solution**
Would it be possible to introduce native support for validating multiple occurrences of values within a single field? For instance, an enhancement to the "enum" constraint that allows defining separators (e.g., ";") or a dedicated "multiEnum" constraint?

If there are alternative approaches within Frictionless that we might have missed, please let us know. Any guidance would be greatly appreciated!

Thank you for your time and consideration.

yarrrly:
- fixes #1741

---

`list` is unchecked on the DPv2 checklist in #1737. On current `main` a schema that uses it is rejected whole:

```python
>>> Schema.from_descriptor({"fields": [{"name": "codes", "type": "list", "itemType": "integer"}]})
FrictionlessException: [field-error] Field is not valid: field type "list" is not supported
```

`validate()` on the same descriptor reports the schema invalid, so a single v2 field makes the whole package invalid.

This adds `ListField` next to the other field types:

- `delimiter` (default `,`, multi-character sequences work) and `itemType` (default `string`, restricted by the profile to the seven types the standard names)
- items are read with the default format of their type, as the standard requires, so `2006-11-21` parses under `itemType: date` and `21/11/2006` does not
- `required`, `minLength`, `maxLength` and `enum`, with length counting items
- writing joins the items back with the delimiter, so a CSV round trip returns the same list

Two decisions worth flagging for review.

An item that fails to cast invalidates the cell (`type is "list/default"`) rather than leaving `None` inside the list, which is what `ArrayField` does. There the JSON array is already parsed and per-item notes make sense; here the lexical form is one string, so a bad item means the string is not a list of that type.

No version gate. `missingValues` needed one in #1787 because the shape of an existing property changed. A new type is an addition, so this follows `fieldsMatch` in #1790 instead.

Left out on purpose: `list` in `DEFAULT_FIELD_CANDIDATES` (any cell with a comma would be detected as a list), parser `supported_types`, and categories on items. #1777 touches `ArrayField` and does not overlap with this.

To try it:

```python
from frictionless import Schema
from frictionless.resources import TableResource

schema = Schema.from_descriptor({"fields": [
    {"name": "id", "type": "integer"},
    {"name": "codes", "type": "list", "itemType": "integer"},
]})
rows = TableResource(data=[["id", "codes"], ["1", "11,22"], ["2", "33"], ["3", "44,bad"]], schema=schema).read_rows()
```

On `main` this raises at `from_descriptor`. Here rows 1 and 2 read as `[11, 22]` and `[33]`, and row 3 gets a type-error on the `codes` cell.

```bash
pytest frictionless/fields/__spec__/test_list.py   # 42 tests
hatch run test
```

I ran the full suite on this branch and on `main` (44661ad) in the same environment and got identical failure sets, all of them from optional extras I don't have installed, plus the 42 new passes. `ruff check frictionless` is clean and `pyright` reports nothing in the new files.


frictionless/fields/__spec__/test_list.py
```diff
@@ -0,0 +1,214 @@
...
+    field = Field.from_descriptor(descriptor)
+    cell = field.read_cell(source)[0]
+    assert cell == target
+
+
+def test_list_read_cell_type_error_note():
+    field = fields.ListField(name="name", item_type="integer")
+    cell, notes = field.read_cell("1,bad,3")
+    assert cell is None
+    assert notes == {"type": 'type is "list/default"'}
```
pierrecamilleri, line 45: This is actually not a very helpful hint, and the error message could be improved. Right now, it is : 

```
Type error in the cell "1,bad,3" in row "1" and field "name" at position "1": type is "list/default"
```

The error does not occur because of the list/default type, but because of the list item_type (actually, the test should be moved below the `# Item Type` title).

A more accurate note would be : 

```
expected item type is "integer"
```

frictionless/fields/__spec__/test_list.py
```diff
@@ -0,0 +1,214 @@
...
+
+def test_list_read_cell_missing_value():
+    field = fields.ListField(name="name")
+    assert field.read_cell("")[0] is None
+
+
+def test_list_read_cell_missing_value_is_not_an_item():
+    field = fields.ListField(name="name", item_type="integer", missing_values=["-"])
+    assert field.read_cell("-")[0] is None
+    assert field.read_cell("1,-,3")[0] is None
```
pierrecamilleri: Ok, but the  tested behavior is not obvious (here "-" is not a valid integer and the cell is read as None).  

I think a more natural test that the missing value placeholder matches the whole string but not a single item would be to have `item_type="string"` and check that  `field.read_cell("1,-,3")[1] == "-"`

frictionless/fields/__spec__/test_list.py
```diff
@@ -0,0 +1,214 @@
...
+    assert field.read_cell("a,b")[1] is None
+    assert field.read_cell("a,b,c")[1] == {"maxLength": 'constraint "maxLength" is "2"'}
+
+
+def test_list_read_cell_enum():
+    field = fields.ListField(
+        name="name", item_type="integer", constraints={"enum": ["1,2"]}
+    )
+    assert field.read_cell("1,2")[1] is None
+    assert field.read_cell("1,3")[1] == {"enum": 'constraint "enum" is "[\'1,2\']"'}
```
pierrecamilleri: I now enum is present in the Array type, but [I do not think it is helpful there in its current form](https://github.com/frictionlessdata/frictionless-py/issues/1718), and it is [not part of the list spec](https://github.com/frictionlessdata/datapackage/blob/da3526c4ed19563620d6a5e33ea55db6234cc411/profiles/dictionary/schema.yaml#L1110).

So please remove the enum constraint for the list type.

frictionless/fields/__spec__/test_list.py
```diff
@@ -0,0 +1,214 @@
...
+        "type": "list",
+        "itemType": "integer",
+        "delimiter": ";",
+    }
+
+
+# Schema
+
+
+def test_list_schema_from_descriptor_issue_1741():
```
pierrecamilleri: issues may be mentioned in tests when there was a regression - not sure it is useful for a new feature like this is the case.

pierrecamilleri (changes requested):
Thank you.

I noted a few possible improvements, and especially, can you please remove the enum constraint.

yarrrly:
Thanks for the review. All four points are in f3fb2e6:

- The enum constraint is gone from `supported_constraints`. A list field declaring `enum` now fails schema validation the same way `pattern` does, and `test_list_read_cell_unsupported_constraint` covers both.
- The item failure note now says `expected item type is "integer"`, and the test sits under `# Item Type`. Cells that are not a string or a sequence at all keep the generic `type is "list/default"` note. The override wraps the default cell reader, same as `ArrayField` does.
- The missing-values test uses `item_type="string"` and asserts `read_cell("1,-,3")[0] == ["1", "-", "3"]`.
- Renamed `test_list_schema_from_descriptor_issue_1741` to `test_list_schema_from_descriptor`.

The end-to-end message now reads: `Type error in the cell "1,bad,3" in row "2" and field "name" at position "1": expected item type is "integer"`.

pierrecamilleri:
Looks good to me ! Thanks

(sorry I closed the PR by mistake)

pierrecamilleri (approved):

